PR #3, opened 2021-07-14
add github actions for building with GCC, Apple Clang, and MSVC

henryborchers:
This currently builds with -Dgearmulator_BUILD_JUCEPLUGIN:BOOL=FALSE.  juce_set_vst2_sdk_path issue is resolved, this line can be taken out


dsp56300 (approved):
I committed that VST2 is skipped if not present, you may omit the skip of the Juce plugin now.

PR #6, opened 2021-07-18
controller dump request

790:
Implements the Controller Dump Request command which sends the requested part patch as CC/Poly messages.


dsp56300 (approved):
thanks for your efforts!

PR #7, opened 2021-07-23
Fix ROM location for macOS standalone.

talaviram:
This allows the rom to be outside the macOS `.app` which is a folder similar to other formats.
Due to codesigning putting the `.bin` in the actual binary in macOS temper with signing and isn't consistent with the other formats.

Notes:
- I've also did a formatting commit using clang-format to avoid mixed tabs/spaces.


dsp56300:
Thanks a lot!

However, seems that I have to update the clang-format (I use Resharper instead) to reflect that I prefer tabs over spaces :)

PR #10, opened 2021-09-25
Fix `Assign3 Source` range and control type.

really-need-an-api-key:
I believe this control should have 28 discrete values instead of being a continuous one.


talaviram:
Thanks. Lgtm

PR #15, opened 2021-09-27
hey hey hey,, namaste

lol-catz:


talaviram (changes requested):
I saw you merge from the main branch. Instead, it's cleaner to rebase your branch / PR this will keep commits ordered better without extra noise.

source/jucePlugin/VirusController.cpp
```diff
@@ -945,7 +945,7 @@ namespace Virus
 
 	juce::String numToKeytrigger(float v, Parameter::Description)
 	{
-		return juce::roundToInt(v) == 0 ? "Off" : "Keytrigger Phase " + juce::String(juce::roundToInt(v));
+		return juce::roundToInt(v) == 0 ? "Off" : juce::String(juce::roundToInt(v));
```
talaviram, line 948: What's the fix here? Remember that exposed parameters are also used for automation lanes.

source/jucePlugin/VirusController.cpp
```diff
@@ -476,7 +476,7 @@ namespace Virus
 		case 5:  return "S&G";
 		case 6:  return "Wave3";
 		case 7:  return "Wave4";
-		case 9:  return "Wave5";
```
talaviram: This commit can be squashed to previous instead. The nice art of have a branch / PR

talaviram (changes requested):

source/jucePlugin/VirusController.cpp
lol-catz: hmm.. I didnt realize that automation issue i just created here..  i dont know the correct answer for this..     the problem with the current  implemtation is the value does not show up in the box next to the slider

source/jucePlugin/VirusController.cpp
lol-catz: i see.  thank you for this info.  immaa git noob... need to study git docs .. sorry for the mess.

source/jucePlugin/VirusController.cpp
talaviram: Start playing with git rebase -i and the oldest commit you'd like to start "rewriting history" for.

Just remember you'll need to force push to this branch afterward as they'll diverge.

lol-catz: noted..  will def be spending some time this week learning to properly work git...  i know there are a lot of gotchas if you don't think things through ahead of time.

lol-catz: back to this specific edit though  what to you propose  the correct solution is?

PR #18, opened 2021-09-30
Fix bank loading order

really-need-an-api-key:
Right now it loads as A,B,C,B,C,D,E,F and Bank C in the popup shows the third read-only bank instead of the first one.
Should be A,B,A,B,C,D,E,F, I think?


dsp56300 (commented):
thanks! Stupid mistake

PR #27, opened 2021-10-04
Add midi/syx bank loading

790:
Works but the code probably needs cleaned up a bit.

### DIFF
--- a/source/jucePlugin/VirusController.cpp
+++ b/source/jucePlugin/VirusController.cpp
@@ -86,7 +86,7 @@ namespace Virus
             }
             else if (i == 5)
             {
-                if (msg[i] != m_deviceId)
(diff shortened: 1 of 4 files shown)


source/jucePlugin/VirusController.cpp
```diff
@@ -86,7 +86,7 @@ namespace Virus
             }
             else if (i == 5)
             {
-                if (msg[i] != m_deviceId)
+                if (msg[i] != m_deviceId && msg[i] != 0x10)
```
dsp56300, line 89: 0x10 could become a constant so it is known what it is

dsp56300 (approved):

dsp56300:
thanks a lot!

PR #30, opened 2021-12-28
add prev/next preset select buttons. expose device id parameter

790:
couldn't find an existing variable that tracked current bank/program per part and the data in the multi didn't seem to be correct so added some new variables to track it's when changed. if there's a better existing place to get that data let me know and i'll update this.

### DIFF
--- a/source/jucePlugin/PluginEditor.cpp
+++ b/source/jucePlugin/PluginEditor.cpp
@@ -68,7 +68,26 @@ AudioPluginAudioProcessorEditor::AudioPluginAudioProcessorEditor(AudioPluginAudi
 			}
 			selector.showMenu(juce::PopupMenu::Options());
 		};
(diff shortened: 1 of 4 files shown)


dsp56300:
Thanks a lot! The microcontroller is aware of the current settings, however, it cannot be accessed. Unfortunately, Access did not come up with any "get current single/multi/part single" MIDI commands so it is impossible to implement "correctly". I need to come up with additional custom midi sysex requests for this I think.

The only issue that I see is that there is no way for the editor to know when a program change via midi modifies the current single/multi, but this also applies to the current dummy UI in various other areas. Even knob changes via midi are not recognized yet, so I think for now this PR is fine. I'm going to test it now.

dsp56300:
works as intended, cool stuff! 👍

PR #31, opened 2022-01-06
Wire up new ui

790:
This also removes the old temporary UI and moves most of it's functions (bank loading etc) into the new one, use 866fe95 to have both available.

most controls that are relevant should work with notable exceptions of arp mode (display), arp hold button (display and function) and the vocoder.

tldr of changes:

Added new `Controller::getParameter` that takes a `part` argument. added `m_part` member to `VirusParameterBinding`. when part changes i set the part on the VPB then recreate the main windows. the .binds pass VPBs current set part to the getParameter call. VPB keeps a list of the VirusParameter.onValueChanged for comboboxes it's has set and removes them when .setPart is called, I couldn't find a simple way to add destroy listener to the combobox, I also wasn't sure if the .referTos need to be cleaned up in some way. There's probably some memory leaks in here.

In `VirusEditor::changePart`, i'm not sure if this is the right way to recreate components but it was one that worked.

`PluginEditor` has been gutted and now it just displays the new UI, it could probably be removed and just have `VirusEditor` be the `AudioPluginAudioProcessorEditor`


dsp56300:
thank you SO MUCH for this!

PR #109, opened 2022-07-18
Page C Undefined76 seems to be the Pure Tuning parameter

samstaton:
I enabled this parameter by giving it a name and making it public. 
I hacked it into the hoverland skin, but feel free to reject that bit if you prefer to keep the right fonts in the skin.


dsp56300:
thanks a lot @samstaton !

PR #130, opened 2022-11-22
Fix: #102, plugin does not recall the user's chosen GUI scale correctly in Bitwig Studio.

### LINKED ISSUE #102: GUI size not remembered in Bitwig
<@517141687743807489>,<#889619013817217034>,yes but doing it every single time i open gui...?,[Jump To](https://discordapp.com/channels/829099347975208970/889619013817217034/986701811115249737)

dreddi:
Fixes #102

### Problem:
In Bitwig Studio, under particular conditions, the VST3 plugin renders at more than twice the size, due to the host calling the plugin's `JuceVST3Editor::setContentScaleFactor` method during plugin setup. Unfortunately, `setContentScaleFactor`, as written, overwrites the user's scaling choice made in the GUI/saved in the config file.

### Affects:
Bitwig Studio users (any DAW that sends `setContentScaleFactor`) -- specifically, Windows users who have display scaling set above 100% and Linux users who use a 4k display resolution.

### Solution:
We patch `JuceVST3EditController::setEditorScaleFactor` to prevent overwriting the user's GUI scale choice. Now, it reads the 'user scale' and calculates the final scale as the 'user scale' times the display's 'content scale factor'. We also patch `AudioProcessorEditor::setScaleFactor` to memorise 'content scale factors' sent by the plugin host (via `JuceVST3Editor::setContentScaleFactor`) in the event the user wishes to change the editor scale manually but not overwrite the 'content scale factor' currently applied.

### DIFF
--- a/source/jucePlugin/PluginEditor.cpp
+++ b/source/jucePlugin/PluginEditor.cpp
@@ -76,7 +76,7 @@ void AudioPluginAudioProcessorEditor::loadSkin(const Skin& _skin)
 
 void AudioPluginAudioProcessorEditor::setGuiScale(int percent)
 {
-	setScaleFactor(static_cast<float>(percent)/100.0f * m_rootScale);
+	setScaleFactor(static_cast<float>(percent)/100.0f * m_rootScale, -1.0);
 	auto* config = processorRef.getController().getConfig();
 	config->setValue("scale", percent);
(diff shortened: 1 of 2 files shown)


dreddi:
Please delete this PR, see https://github.com/dsp56300/gearmulator/pull/131

PR #134, opened 2022-11-28
fixes #132: non-GUI triggered parameter changes now update the GUI control label

### LINKED ISSUE #132: <@928936217117081611>,<#841809045505703956>,i have an issue, regarding the visual fb of the values in the "display". i think its maybe in the 
programming - there is some missing class link or smthing.

if you turn a knob or change a parameter the display clearly shows the value and give an feedback.
but if the Osirius is controlled via an externel hardware eg MidiController - for example "CC17 waveshape"
the knob is turning and the value changes, but there is no visual feedback in the display.

so its pretty hard to set for example the correct semitones in a second osc or such stuff...

so if there is a visible value output in the display when the mouse rotates the knob why is there no 
value output when the CC rotates the same knob ?,[Jump To](https://discordapp.com/channels/829099347975208970/841809045505703956/1044562443285827594)
<@928936217117081611>,<#841809045505703956>,i have an issue, regarding the visual fb of the values in the "display". i think its maybe in the 
programming - there is some missing class link or smthing.

if you turn a knob or change a parameter the display clearly shows the value and give an feedback.
but if the Osirius is controlled via an externel hardware eg MidiController - for example "CC17 waveshape"
the knob is turning and the value changes, but there is no visual feedback in the display.

so its pretty hard to set for example the correct semitones in a second osc or such stuff...

so if there is a visible value output in the display when the mouse rotates the knob why is there no 
value output when the CC rotates the same knob ?,[Jump To](https://discordapp.com/channels/829099347975208970/841809045505703956/1044562443285827594)

dreddi:
Fixes #132 

## Solution:

We patch `Virus::Controller::sendParameterChange` to update the GUI control label on every parameter change. 

This is achieved by implementing a new `VirusEditor::updateControlLabel`  that takes a `pluginLib::Parameter` as an argument, and updates the control label using its fields. We refactor the existing code to ensure that `FocusedParameterTooltip` behaviour continues as before.

Examples:
1. In single mode, updating 'Portamento Time' via midi controller or DAW plugin wrapper will display 'Portamento Time' in the control label.
2. In multi mode, you get the same behaviour, but also
3. if channel 5 is the currently active part, updating channel 2's 'Portamento Time' will display 'Ch 2 Portamento Time' rather than 'Portamento Time', to reduce ambiguity.

Other changes:
1. [source/juceUiLib/tabgroup.h:](https://github.com/dsp56300/gearmulator/pull/134/files#diff-209e655b6314d6bdc2810e6efcf9a737912eff9c876a79c4f033489194c5cad8R425) We introduce a unique ID for each `juce::Component` during `template <typename T> T* UiObject::createJuceObject(Editor& _editor, T* _object)` , so that we can match on this key by calling `component.getComponentID()`, for example. We use this when checking whether a component exists on a particular page (for `FocusedParameterTooltip`). 

Possible improvements:

- ~Investigate any performance implications, of walking the `m_bindings` vector linearly, on every single parameter change.~ This code has been made redundant.
- A small number of parameters will not trigger a message in the 'control label' (Ch1 Channel Volume, Ch1 Balance, possibly others). This occurs when the skin does not implement controls for these parameters. We could choose to implement labels for these parameters in a fall back capacity (by referencing the labels given in `parameterDescriptions_C.json`.
  - These parameters can always be reached from a DAWs plugin wrapper UI, or through a DAWs automation panel, and so a missing control label for them may be noticed, but irrelevant to the user's workflow.

### DIFF
--- a/source/jucePlugin/PluginEditor.cpp
+++ b/source/jucePlugin/PluginEditor.cpp
@@ -260,3 +260,7 @@ void AudioPluginAudioProcessorEditor::mouseDown(const juce::MouseEvent& event)
 
 	openMenu();
 }
+
+const juce::Component * AudioPluginAudioProcessorEditor::getVirusEditor() const {
+	return m_virusEditor.get();
+}
--- a/source/jucePlugin/PluginEditor.h
+++ b/source/jucePlugin/PluginEditor.h
@@ -13,7 +13,8 @@ public:
 
     void paint (juce::Graphics&) override;
     void resized() override;
-	void mouseDown(const juce::MouseEvent& event) override;
+    void mouseDown(const juce::MouseEvent& event) override;
+    const juce::Component * getVirusEditor() const;
 
 private:
 	struct Skin
--- a/source/jucePlugin/VirusController.cpp
+++ b/source/jucePlugin/VirusController.cpp
@@ -4,6 +4,7 @@
 
 #include "ParameterNames.h"
 #include "PluginProcessor.h"
+#include "PluginEditor.h"
 
 #include "../virusLib/microcontrollerTypes.h"
 #include "../synthLib/os.h"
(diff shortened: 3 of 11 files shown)


dreddi:
Bug: control label only updates correctly for the channel selected in the GUI (Ch1 in single mode, the selected channel in multi mode). Control label doesn't update when a different channel's value is changed. Will fix.

dreddi:
Bugs fixed, ready to merge. Original post has been updated.

dreddi:
Ran `clang-format` on the lines of code I touched, using `source/.clang-format` as a ruleset

dsp56300:
thanks for the suggestion, I added it meanwhile

PR #148, opened 2023-01-10
build(git): convert submodule from relative to absolute link

grawlinson:
Git complains that the submodule cannot be found, and fails to fetch it.


dsp56300 (approved):
ouch, thanks

PR #171, opened 2023-05-15
Add several more zoom levels

mkruselj:
Also update gitignore to filter out /.vscode and /build/ folders


dsp56300:
appreciated, thanks!

PR #187, opened 2024-05-08
Capitalize Rom to ROM in the browser for Os(t)irus

mkruselj:


dsp56300:
Sorry but I have to reject this for now, the names are used as data source names and if they are modified, people lose their tags & favourites. I need to come up with a better solution but for now this doesn't work

mkruselj:
Bank names could be considered as non-case sensitive perhaps?

PR #190, opened 2024-05-09
Filter Extra knob changes label name contextually

mkruselj:
Also, renamed that parameter from F1ContextSensitive to F1Extra, since that's how it's called on hardware. We can afford to do these breaking changes while we are in alpha 🙂 

Showcase:

context final


dsp56300:
Nice. I will need some time to recreate this in Unity though :)

mkruselj:
@dsp56300 Unity the game engine? 😮

PR #195, opened 2024-05-11
Improve column breaks for dropdown menu based parameters

mkruselj:
Previously column breaks happened after every 16th menu entry verbatim, no chance of parole. With this change a column break will only happen if we have more than one and a half columns of entries total (16 + 8).

This allows changing menus like part MIDI channel in Xenia from looking like this:

image

into looking like this:

image

which is far more visually agreeable, I think 🙂

### DIFF
--- a/source/jucePluginLib/parameterbinding.cpp
+++ b/source/jucePluginLib/parameterbinding.cpp
@@ -133,10 +133,15 @@ namespace pluginLib
 	
 		uint32_t count = 0;
 
-		for (const auto& vs : sortedValues)
+		// we want our long menus to be split into columns of 16 rows each
+		// but only if we have have more entries than one and a half such column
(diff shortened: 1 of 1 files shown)


source/jucePluginLib/parameterbinding.cpp
```diff
@@ -133,10 +133,15 @@ namespace pluginLib
 	
 		uint32_t count = 0;
 
-		for (const auto& vs : sortedValues)
+		// we want our long menus to be split into columns of 16 rows each
+		// but only if we have have more entries than one and a half such column
+		const uint32_t splitAt = (sortedValues.size() > 24) ? 16 : 0;
+
+		for (const auto &vs : sortedValues)
```
dsp56300, line 140: minor: & should be next to auto as it is part of the type

dsp56300 (approved):
thanks!

PR #196, opened 2024-05-12
Constrain tooltip to plugin window

mkruselj:
Unsure if there is a better way, but for now this seems to work pretty nicely!

### DIFF
--- a/source/jucePluginEditorLib/focusedParameterTooltip.cpp
+++ b/source/jucePluginEditorLib/focusedParameterTooltip.cpp
@@ -1,10 +1,13 @@
 #include "focusedParameterTooltip.h"
 
+#include "pluginEditor.h"
+
 #include "juce_gui_basics/juce_gui_basics.h"
 
 namespace jucePluginEditorLib
 {
-	FocusedParameterTooltip::FocusedParameterTooltip(juce::Label* _label) : m_label(_label), m_defaultWidth(_label ? _label->getWidth() : 0)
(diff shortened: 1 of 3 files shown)


source/jucePluginEditorLib/focusedParameterTooltip.cpp
```diff
@@ -46,22 +49,16 @@ namespace jucePluginEditorLib
...
-			}
-			*/
 			if(m_label->getProperties().contains("offsetY"))
 				y += static_cast<int>(m_label->getProperties()["offsetY"]);
 
 			m_label->setTopLeftPosition(x,y);
 			m_label->setSize(labelWidth, m_label->getHeight());
+
+			const auto editorRect = m_editor.getBounds();
+			auto labelRect = m_label->getBounds();
```
dsp56300, line 59: can be const too

source/jucePluginEditorLib/focusedParameterTooltip.h
```diff
@@ -21,5 +26,6 @@ namespace jucePluginEditorLib
 	private:
 		juce::Label* m_label = nullptr;
 		int m_defaultWidth;
+		const genericUI::Editor& m_editor;
```
dsp56300, line 29: can be of type juce::Component an could be named m_bounds

dsp56300 (approved):

dsp56300:
thanks!

PR #198, opened 2024-05-13
Add "displayTime" property to the tooltip component in the skin engine

mkruselj:
Also reduce the default tooltip time to 2 seconds. I prefer it even shorter (500 ms) but now I can set it up for myself. 🙂 3 seconds definitely felt too long IMO!


dsp56300:
thanks a lot!

PR #201, opened 2024-05-30
More Xenia fixes

mkruselj:
Extend parameter range for mod destinations to include F1 Extra, FM Amount and Mod 2-4 Amount
Remap mod destinations for more logical order
Remap Play Parameter list for more logical order
Remap Wavetable list to remove <reserved> slots

We're still alpha so breaking changes like modifying WT list is still good :)


dsp56300 (commented):
thanks!

PR #208, opened 2024-06-08
Xenia: create new parameter for LFO clocked mode

mkruselj:
This will allow us to show actual note values in the tooltip!


dsp56300:
thanks 👍

PR #220, opened 2024-08-25
Cleaner looking skin submenu

mkruselj:
The user doesn't really need to know the file extension, and I would say that the relative path to skin is also not really important. It would probably be better to add an option to reveal the currently selected skin in Explorer/Finder instead (in a separate PR).

image


beely:
Yes, *please* do this - we already know the directory the skins are coming up in because we would have had to put them there in the first place - and we don't really case the file format extension... ;)

PR #228, opened 2024-11-03
Better folder locations for settings on Linux

foresense:
- They used to clutter the home folder. Moving to ~/.config/The Usual Suspects/*
- In line with the folder locations for roms and skins that are in ~/.local/share/The Usual Suspects/*


dsp56300:
Sorry for rejecting it but there seems to be a misunderstanding here: These are the old locations and are kept for backwards compatibility. If data in the old folders is found, it is migrated to the new locations, these are defined by `synthLib::getSpecialFolderPath()`, if the old paths are altered this breaks compatibility with old installations

foresense:
Oh oops, my bad! Thanks for fixing, I'll update to master for now then.

PR #251, opened 2025-12-28
fix missing cstdint includes

ipatix:
The headers in the affected files are missing the cstdint headers. On Debian trixie this causes compile errors.

Should fix #248


dsp56300 (approved):
thanks!

PR #264, opened 2026-04-26
Sync patchmanager UI selection with host-driven setStateInformation

AxidentDK:
## Summary

When a VST3 host (e.g. Native Instruments Komplete Kontrol loading an NKS preset) calls `setStateInformation`, the audio engine correctly restores the new patch via `Plugin::setState`, but the patchmanager UI keeps showing whatever was previously selected (or nothing). The selection state is set only from manual UI clicks via `setSelectedPatch`, never from the state-restore path.

This PR fixes that with a small generic mechanism that benefits every gearmulator-based plugin (Microwave XT, Virus, microQ, JD-800, Nord Lead 2X) without per-plugin code changes.

## Implementation

- New parameterless event `Processor::evStateLoaded` (jucePluginLib).
- Fired at the end of `Processor::setState`, on both the synchronous and async device-recreate paths.
- `PatchManager` subscribes in its constructor and, on the message thread, hashes whatever the audio engine is currently playing on the active part (via the existing `requestPatchForPart` convenience overload, which already computes the per-plugin MD5 hash) and then linear-scans loaded data sources for a matching content hash.
- A match drives `setSelectedPatch`. No match leaves selection unchanged (clearing it would surprise users who manually loaded one-off SysEx imports).

The hash-lookup approach means no per-plugin code change is required — each subclass already implements `requestPatchForPart` and `initializePatch` which together provide the canonical hash for the active patch.

## How was this discovered

While building Automated Preview Creator (APC), a tool that generates NKS preview files for VST3 plugins so musicians can browse and load presets directly from NI hardware. APC's generated NKS files correctly cause the audio engine to play the right patch (verified with both setstate-getstate round-trip and live audio rendering), but the plugin's patchmanager UI stayed empty after KK loaded the preset. Investigation traced this to `setSelectedPatch` only being reachable from UI click paths, never from `setStateInformation`.

## Test plan

- Build with each plugin variant (XT, Virus, microQ, JD-800, NL2X).
- Manual: load a known-cached patch via NKS / external host setStateInformation; patchmanager UI should highlight the matching entry.
- Manual: load a foreign SysEx (no match in cache); patchmanager UI selection should remain unchanged (no false clear).
- Manual: ensure manual-click selection still works as before (regression check).
- Verify no thread issues — the listener defers to the JUCE message thread via `juce::MessageManager::callAsync` and re-checks the instance is still alive (same pattern used by the existing program-change router subscription).

### DIFF
--- a/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
+++ b/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
@@ -71,6 +71,22 @@ namespace jucePluginEditorLib::patchManager
 			});
 
 		setUi(std::make_unique<patchManagerRml::PatchManagerUiRml>(m_editor, *this, *m_editor.getRmlComponent(), _rootElement, *m_editor.getPatchManagerDataModel(), _groupTypes));
+
+		// Subscribe to host-driven state loads. Defer the work onto the JUCE
+		// message thread because evStateLoaded may fire from the audio thread
+		// (or any host thread that drove setStateInformation), and the patch
+		// search + UI update must run on the message thread.
+		m_onStateLoaded.set(_editor.getProcessor().evStateLoaded, [this]
+		{
+			auto* self = this;
+			juce::MessageManager::callAsync([self]
+			{
+				std::lock_guard lock(getInstancesMutex());
+				if (getInstances().find(self) == getInstances().end())
+					return;
+				self->selectActivePatch(self->getCurrentPart());
+			});
+		});
 	}
 
 	PatchManager::~PatchManager()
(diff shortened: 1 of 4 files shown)


dsp56300:
First: apologies this sat since April without a reply. Keeping it open — the problem you are solving is real and worth fixing, and the diff is properly scoped to one concern, which we appreciate. A few things need addressing before we can merge it.

**What is right.** The publish side is done the way we want it: firing `evStateLoaded` and hopping to the message thread through the existing static instance-set guard, matching the `onProgramChangeQueued` handler a few lines above it in the same constructor. Good threading discipline, and the use case (host-driven state restore, e.g. NKS/Komplete Kontrol) is one we care about.

**1. The consume side reimplements something we already have.** `selectActivePatch()` hand-rolls a linear scan:

```cpp
std::vector<pluginLib::patchDB::DataSourceNodePtr> sources;
getDataSources(sources);
for (const auto& ds : sources)
    for (const auto& p : ds->patches)
        if (p && p->hash == active->hash) { setSelectedPatch(_part, p); return; }
```

That is what `updateStateAsync()` / `DB::findDatasourceForPatch()` already do, and they do it better. In most cases this should collapse to:

```cpp
updateStateAsync(_part, requestPatchForPart(_part));
```

**2. The duplication introduces a data race, which is the blocking issue.** The existing search path runs on the loader thread, which serialises naturally against the loader thread's own unsynchronised writes to `ds->patches` in `DB::addDataSource()`. Your copy runs on the **message thread** inside `callAsync` and iterates `ds->patches` with nothing held — `getDataSources()` takes a `shared_lock` only long enough to copy the pointers, it does not protect the per-node set. A host state-load arriving while the initial background scan is still running is a realistic scenario, not a theoretical one. Reusing `updateStateAsync` avoids this by construction.

**3. Weaker matching.** You compare `p->hash == active->hash` only, dropping the sysex-size + name fallback (`equals()`) that the existing search applies for dumps that are not byte-identical.

**4. Only the current part is resynced.** `selectActivePatch(getCurrentPart())` leaves the other parts stale on multitimbral synths; `onLoadFinished()` loops over every part for this reason.

If you rework it around `updateStateAsync`, points 2–4 all fall away together. Happy to look at it again after that.

dsp56300:
Heads-up: the source tree was restructured by ownership and manufacturer in `275285257`, which is now on `main`, so this PR conflicts. That is unrelated to the review feedback above — it affects every open PR that touches `source/`.

| old | new |
|---|---|
| `source/jucePluginEditorLib/patchmanager/patchmanager.cpp/.h` | `source/framework/juce/jucePluginEditorLib/patchmanager/…` |
| `source/jucePluginLib/processor.cpp/.h` | `source/framework/juce/jucePluginLib/processor.cpp/.h` |

There is no blanket `source/` include root any more, but every library exports its own parent as its include root, so `#include "jucePluginLib/processor.h"` and similar keep working unchanged. Only depth-relative includes (`../../`) would break.

The review points from my previous comment still stand — in particular reusing `updateStateAsync()` instead of the hand-rolled scan, which also removes the message-thread data race on `ds->patches`. If you rework it that way you will be touching those files anyway, so the rebase and the fix can land in one go.

PR #265, opened 2026-06-01
Feature/rmlui svg plugin

kartun83:
# Add RmlUi SVG plugin support (LunaSVG) and fix macOS build
## Summary
Adds LunaSVG as a bundled third-party submodule and enables the RmlUi SVG plugin, allowing skins to render SVG graphics — both from external files and inline in RML. Also fixes two pre-existing macOS build issues discovered during this work.

Depends on: dsp56300/RmlUi#1 (SVG plugin CoreInstance API fixes + inline SVG support).

### SVG plugin integration
New submodule: source/3rdparty/lunasvg — LunaSVG v3.5.0, **pinned**. Includes its plutovg rendering backend. Built as a static library with examples disabled.

source/3rdparty/CMakeLists.txt: adds add_subdirectory(lunasvg) before RmlUi so the lunasvg CMake target exists when RmlUi's Dependencies.cmake searches for it. Sets RMLUI_SVG_PLUGIN=ON.

RmlUi's own find module detects the lunasvg target automatically via add_subdirectory — no manual path configuration needed.

Once built, skins can use SVG in two ways:

```svg
<!-- external file -->
<svg src="envelope.svg" class="jucePos" style="width: 200dp; height: 80dp;"/>
```

```svg
<!-- inline — no file needed -->
<svg xmlns="http://www.w3.org/2000/svg" class="jucePos"
     viewBox="0 0 200 80" style="width: 400dp; height: 160dp;">
  <polyline points="0,80 20,10 60,50 120,50 180,80"
            stroke="#e8650a" stroke-width="2" fill="none"/>
</svg>
```

And Lua scripts can update SVG content at runtime for parameter-driven graphics (envelopes, filter curves, etc.) — see below for details.

## What this enables

Together, inline SVG and `params.onChange` turn the skin layer into a **live drawing surface** — not just a layout of predefined widgets, but a canvas where anything SVG can describe can be rendered, updated, and driven by synth parameters in real time.

### Parameter-driven visualizations

Displays that were previously impossible without a dedicated C++ component or a library of pre-rendered PNG sprites:

- **Envelope displays** — ADSR/ADSSR silhouette redraws as any knob turns;   bezier curves give smooth, natural shapes
- **Modulation depth arcs** — a colored arc around a knob showing the   current modulation source, depth, and direction (like Serum's mod ring),  one `<svg>` overlay per target parameter
- **LFO waveform preview** — shape switches from sine to saw to square as  the LFO Shape combo changes
- **Filter response sketch** — approximate resonance peak and roll-off curve  drawn from Cutoff and Resonance values

### Custom graphics without raster assets

Every shape SVG can express — paths, bezier curves, arcs, polygons, gradients, masks — is available with no pre-rendered PNG.  A skin that previously needed a 128-frame knob sprite atlas for a custom indicator can replace it with a few lines of SVG and Lua.  The result is:

- **Resolution-independent** — scales cleanly at any DPI without artefacts
  from upscaled bitmaps
- **Trivially recolorable** — change a hex value in Lua, not 128 PNG frames
- **Zero art pipeline** — iterate on shapes in code, not in an image editor

### Escaping the RML element library

RML provides a fixed set of interactive elements — knobs, buttons, combos, sliders.  SVG breaks out of that constraint for the *display* layer.
Anything that needs to look custom but does not itself need to be interactive (indicators, meters, curves, annotations, section dividers with complex geometry) can be drawn entirely in SVG and driven by Lua.

The practical pattern is a layered composition:

```
[SVG layer]  — draws the visual: curve, arc, shape   (non-interactive)
[RML layer]  — transparent interactive elements above: knob, button
```

The user interacts with the RML element; the SVG responds via `params.onChange`.  This is how the envelope display and modulation arcs work — the knobs remain standard JUCE controls, but the SVG beneath them provides context that no static sprite could.

### What it does not replace

SVG cannot receive input events (drag, click) — interactive controls still need RML/JUCE elements.  For very high-frequency updates (e.g. a VU meter at audio rate) the re-rasterization cost would be too high; this approach is suited to parameter-change rates, not audio-rate display.

---

## 1. Inline SVG rendering

The `<svg>` element works without a `src` file — inline SVG markup is rendered directly by the LunaSVG plugin.  Any valid SVG content works: paths, polylines, circles, text, gradients.

```xml
<!-- static inline envelope shape — no external file needed -->
<svg xmlns="http://www.w3.org/2000/svg" class="jucePos"
     viewBox="0 0 200 80"
     style="left: 100dp; top: 200dp; width: 400dp; height: 160dp;">
  <polyline points="0,80 20,10 60,50 120,50 180,80"
            stroke="#e8650a" stroke-width="2" fill="none"/>
</svg>
```

**Note on sizing:** `viewBox` should match the *rendered* pixel dimensions (dp × rootScale).  At the default `rootScale="0.500000"`, a `400dp × 160dp` element renders at `200 × 80 px`, so `viewBox="0 0 200 80"` makes SVG
coordinates equal rendered pixels — convenient for Lua geometry math.

---

## 2. Dynamic SVG via Lua

The `<svg>` element's content can be replaced at runtime using `inner_rml`. Combined with `params.onChange`, this enables fully parameter-driven visualizations — envelope shapes, modulation arcs, waveform displays — without
any C++ changes.

### Requirements

- The element **must use an explicit closing tag** (`></svg>`, not `/>`) for `inner_rml` to work.
- `params.onChange` returns a listener ID.  Store it and call `params.removeListener(id)` on re-init to prevent listener  accumulation across document reloads.

### RML

```xml
<!-- empty element — Lua injects content at runtime -->
<svg xmlns="http://www.w3.org/2000/svg" id="my_display" class="jucePos"
     viewBox="0 0 200 80"
     style="left: 100dp; top: 200dp; width: 400dp; height: 160dp;"></svg>
```

### Lua (`myskin.lua`)

```lua
local _display_lid = nil   -- listener ID for cleanup on reload

local function redrawDisplay()
  local el = document:GetElementById("my_display")
  if not el then return end

  -- Read a parameter (0–127)
  local value = params.get("Cutoff") or 0
  local t     = value / 127.0           -- normalised 0..1

  -- Map to SVG coordinates (viewBox 0 0 200 80)
  local x = math.floor(t * 180) + 10   -- endpoint x, 10..190
  el.inner_rml =
    string.format('<polyline points="0,80 20,10 60,50 %d,50 180,80"'..
      ' stroke="#e8650a" stroke-width="2" fill="none"/>', x)
end

function initDisplay()
  -- Clean up previous listener (safe to call on every document reload)
  if _display_lid ~= nil then
    params.removeListener(_display_lid)
    _display_lid = nil
  end

  redrawDisplay()   -- initial draw

  _display_lid = params.onChange("Cutoff", function()
    redrawDisplay()
  end)
end
```

Call `initDisplay()` from the `<body>` `onload` handler:

```xml
<body onload="initDisplay()" ...>
```

---

## macOS build fixes (unrelated to SVG, discovered during bringup)

build_mac.sh: the script pointed to ./temp/cmake but the configured output directory is temp/cmake_macos. Also called cpack as a bare command, which is not on PATH when CMake is installed as CMake.app. Both fixed.

source/juceRmlUi/CMakeLists.txt: the SSE flag pairs for macOS universal builds were broken. CMake was deduplicating the second -Xarch_x86_64, leaving a bare -msse4.1 that clang rejects on the arm64 slice:

clang: error: unsupported option '-msse4.1' for target 'arm64-apple-macos10.13'
Fixed by wrapping each -Xarch_x86_64/-mflag pair with SHELL: to prevent deduplication:

_before — CMake deduplicates -Xarch_x86_64, arm64 slice gets bare -msse4.1_
target_compile_options(juceRmlUi PRIVATE -Xarch_x86_64 -mssse3 -Xarch_x86_64 -msse4.1)

_after — each pair is preserved as a single argument_
target_compile_options(juceRmlUi PRIVATE "SHELL:-Xarch_x86_64 -mssse3" "SHELL:-Xarch_x86_64 -msse4.1")
Files changed
File	Change
.gitmodules	Register source/3rdparty/lunasvg submodule
source/3rdparty/lunasvg	LunaSVG v3.5.0 submodule
source/3rdparty/CMakeLists.txt	add_subdirectory(lunasvg), RMLUI_SVG_PLUGIN=ON
source/3rdparty/RmlUi	Submodule pointer updated to SVG-plugin-fixed commit
build_mac.sh	Correct output directory, fix cpack path
source/juceRmlUi/CMakeLists.txt	SHELL: prefix for SSE flag pairs

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
+
 # Enable RmlUi Lua bindings and point to our bundled Lua
 set(RMLUI_LUA_BINDINGS ON CACHE BOOL "" FORCE)
 set(RMLUI_LUA_BINDINGS_LIBRARY "lua_as_cxx" CACHE STRING "" FORCE)
+set(RMLUI_SVG_PLUGIN ON CACHE BOOL "" FORCE)
 
 # Create the Lua::Lua target so RmlUi's Dependencies.cmake finds it and skips find_package(Lua)
 add_library(Lua::Lua INTERFACE IMPORTED)
--- a/source/juceRmlUi/CMakeLists.txt
+++ b/source/juceRmlUi/CMakeLists.txt
@@ -78,7 +78,7 @@ endif()
 if(APPLE)
     # Universal or x86_64-only: scope flags to the x86_64 slice
     if(CMAKE_OSX_ARCHITECTURES MATCHES "x86_64" OR CMAKE_SYSTEM_PROCESSOR MATCHES "x86|X86|amd64|AMD64")
-        target_compile_options(juceRmlUi PRIVATE -Xarch_x86_64 -mssse3 -Xarch_x86_64 -msse4.1)
+        target_compile_options(juceRmlUi PRIVATE "SHELL:-Xarch_x86_64 -mssse3" "SHELL:-Xarch_x86_64 -msse4.1")
     endif()
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "x86|X86|amd64|AMD64|i[3-6]86")
     include(CheckCXXCompilerFlag)


dsp56300:
Thank you for this, and apologies that it sat here for so long without a response.

SVG support has since been implemented independently on `main` (commit `99f315860`, 2026-07-23) using the same approach you proposed here — lunasvg as a submodule plus `RMLUI_SVG_PLUGIN` — so the substance of this PR is already in the tree. Closing as superseded. That the two implementations landed on the same design is a good sign your call was the right one; the delay here is on us, not on the PR.

One part of this PR is **not** in `main`, and it is a real fix we are taking separately with credit to you:

```cmake
-target_compile_options(juceRmlUi PRIVATE -Xarch_x86_64 -mssse3 -Xarch_x86_64 -msse4.1)
+target_compile_options(juceRmlUi PRIVATE "SHELL:-Xarch_x86_64 -mssse3" "SHELL:-Xarch_x86_64 -msse4.1")
```

CMake de-duplicates compile options, so the repeated `-Xarch_x86_64` collapses and `-msse4.1` leaks onto the arm64 slice of universal macOS builds. `SHELL:` is the documented way to keep paired flags together. Good catch — that one is going in on its own.

Issue #241, opened 2025-05-22
OsTirus - ARP sync problem in FL Studio 24


paulrigel:
Every long MIDI note which in Ostirus uses arpeggiator starts to play out of sync after few notes.

I saw here a similar issue with Reaper.
This has probably the same root problem - syncing the events with the Main Tempo in ARP Mode.

I'm on Mac.

Such a setting causes trouble: 

![Image](https://github.com/user-attachments/assets/a95d8cb2-65b7-4c3b-87c1-3e6d4f7bd670)

whitespace-rebel:
Same issue in Renoise on Linux.
When starting playback with say 95 BPM, the ARP tempo is wrong at the beginning, but catches up after one or two bars. Probably 120 BPM is hardcoded somewhere, because a project with 120 BPM sounds mostly in sync.

kartun83:
mind testing ?
https://drive.google.com/drive/folders/1G-VhNtk0lnJk6L0JBRThI0qtsNNdA9Ri?usp=sharing
Sorry, mac only :)

PR #266, opened 2026-06-01
Fix/arp clock sync

### LINKED ISSUE #205: Reaper (cokos): Lots of stucked notes and problems Syncing Arpeggiator
Hi!

If I don't start playing the song exactly  on the start of mesaure or its divisions, the arpeggio is out of sync.
Stuck notes almost always happen all the time.

https://www.dropbox.com/scl/fi/sxk3m8fxxaix2bwkrf3hh/VID_20240607_081757.mp4?rlkey=0aogh3ge3dv59rot62540vw3q&st=jws90pak&dl=0

Thx!

kartun83:
Fix MIDI clock phase and loop re-sync for ARP sync issues

Fixes #205, #241

Two bugs in `synthLib/MidiClock` caused the arpeggiator to play out of phase with the host tempo in all DAWs.

---

## Bug 1 — wrong initial clock phase in `start()`

The formula used to initialise `m_clockTickPos` on playback start:
```c++
    quarterPos     = (ppqPos - floor(ppqPos + 1.0))
    m_clockTickPos = quarterPos * 24
```

produces values in `[-24, 0)`. The tick-generation loop only fires a tick when `m_clockTickPos` crosses 0, so starting at `-24` suppresses all ticks for 24 full tick periods — one entire beat at any tempo.

This is exactly the symptom reported: *"ARP tempo is wrong at the beginning, but catches up after one or two bars."* The "catch up" was the clock finally escaping its one-beat head start.

**Fix** — keep `m_clockTickPos` in `[-1, 0)` by computing the fractional position within the *current* tick period:

```c++
    absoluteTicks  = ppqPos × 24
    fracTick       = absoluteTicks − floor(absoluteTicks)
    m_clockTickPos = fracTick − 1.0
```

| ppqPos | fracTick | m_clockTickPos | first tick fires after |
|--------|----------|----------------|----------------------|
| 0.0    | 0.0      | −1.0           | 1 full tick period   |
| 0.5    | 0.0      | −1.0           | 1 full tick period (on a tick boundary) |
| 0.52   | 0.48     | −0.52          | 0.52 of a tick period |

This also fixes the issue reported in #205 where starting playback mid-beat caused immediate de-sync: the old formula ignored the sub-beat position entirely, the new one accounts for it precisely.

---

## Bug 2 — loop and backward-seek not detected

`process()` only called `start()` on the `false → true` transition of `_isPlaying`. When the DAW looped back to an earlier bar, `ppqPos` jumped backward while `m_isPlaying` remained `true`, so `start()` was never called and the clock continued accumulating at the wrong phase indefinitely.

**Fix** — track `m_lastPpqPos` and restart the clock whenever `ppqPos` drops by more than 0.5 beats during playback. The 0.5-beat threshold is safely above any normal per-buffer advance (e.g. 300 BPM with a 4096-sample buffer ≈ 0.28 beats) while reliably detecting all practical loop-back distances.

---

Both fixes are in `source/synthLib/midiClock.cpp` and affect all synths that use the shared MIDI clock (Osirus, OsTIrus, Vavra, Xenia, etc.), but the ARP sync issue was primarily reported against OsTIrus.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
+
 # Enable RmlUi Lua bindings and point to our bundled Lua
 set(RMLUI_LUA_BINDINGS ON CACHE BOOL "" FORCE)
 set(RMLUI_LUA_BINDINGS_LIBRARY "lua_as_cxx" CACHE STRING "" FORCE)
+set(RMLUI_SVG_PLUGIN ON CACHE BOOL "" FORCE)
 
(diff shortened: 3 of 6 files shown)


dsp56300:
**The fix in this PR is correct and we want it.** Closing only because of how the branch is packaged — please read to the end.

We verified the phase math by hand. The old `start()` computed `quarterPos = ppqPos - floor(ppqPos+1.0)`, which for `ppqPos = 2.3` yields `m_clockTickPos = -16.8`. Since ticks only fire once the accumulator crosses zero, the first MIDI clock tick was delayed by up to 24 ticks — i.e. an arp or synced LFO stays silent for up to a full quarter note when playback starts mid-beat. Your replacement computes the fractional distance to the next 1/24 boundary correctly (`-0.8` for the same input), and the loop-back resync branch covers the dominant real-world case of editing under a looped region. Nicely done.

The problem is that this branch is **stacked on `feature/rmlui-svg-plugin`** rather than on `main`. The diff therefore also carries `.gitmodules`, `build_mac.sh`, the RmlUi submodule bump and the new lunasvg submodule — none of which belong to an arp clock fix. That is why GitHub reports this as CONFLICTING, and it is the same reason several of your other PRs conflict: they form a chain, so each one re-contains everything before it.

**Please resubmit this rebased on current `main`, containing only `source/synthLib/midiClock.cpp` and `midiClock.h`.** We will merge it.

One optional follow-up, not a blocker: only backward jumps are resynced, so a *forward* seek during playback still leaves the tick phase stale. That case was equally unhandled before your PR, so it is not a regression.

PR #267, opened 2026-06-01
CLAP: Fix tail length for Virus 

kartun83:
## Problem

When bouncing or freezing a track in a CLAP host (Bitwig, etc.), the host asks the plugin how long it should keep processing audio after the last note-off. `getTailLengthSeconds()` returned `0.0`, so the host
stopped the audio engine immediately at the end of the MIDI clip. Reverb decay and delay echoes produced by the hardware after notes stop were cut off in the rendered file.

This does not affect live playback — the host keeps the engine running regardless. It only matters for offline render / bounce / freeze, where the host needs a definitive "when can I stop?" answer.

## Fix

`VirusProcessor::getTailLengthSeconds()` now reads the active patch parameters and returns the tail the hardware would actually produce.

Reverb tail — Reverb Time (page 110, index 4, range 0–127) is mapped linearly to 0–16 seconds, matching the Virus TI hardware's documented maximum reverb decay time. Returns 0 when Reverb Send is 0: no signal reaches the reverb engine on the hardware either, so there is no tail.

Delay tail — computed as delayTime / (1 − feedback), capped at 30 s to avoid the singularity at full feedback. Delay Time 0–127 maps to 0–1000 ms in free-running mode. Returns 0 when Delay Send is 0.

A dry patch with both sends at zero returns 0 and the host behaves identically to before. A pad with Reverb Time at maximum extends the render by up to 16 seconds, allowing the full decay to be captured.

## Supporting changes

- `getControllerConst()` added to `pluginLib::Processor` as a protected const accessor that returns the controller pointer without  triggering lazy creation. Required to read parameters from a  `const`-qualified override.
- Reverb/delay parameter name constants added to `ParameterNames.h`.

## Scope

Applies to Osirus (Virus A/B/C) and OsTIrus (Virus TI/TI2/Snow) via the shared `VirusProcessor`. Other synths continue returning 0.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
+
 # Enable RmlUi Lua bindings and point to our bundled Lua
 set(RMLUI_LUA_BINDINGS ON CACHE BOOL "" FORCE)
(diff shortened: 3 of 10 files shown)


dsp56300:
Closing for two reasons — one packaging, one substantive.

**1. The fix does not do what the title says.** It reports a tail for Virus TI, but it lives in the shared `virusJucePlugin/VirusProcessor.cpp`, which also backs classic Virus A/B/C (Osirus). Classic Virus has no `Reverb Send` and no `Delay Send` parameter at all — see `source/osirusJucePlugin/parameterDescriptions_C.json`, which only has a combined `Delay/Reverb Mode` selector plus `Delay Time` / `Delay Feedback` / `Reverb Damping`. Both branches here are gated on the Send parameter existing:

```cpp
if(reverbSend && reverbTime && ...)
if(delaySend && delayTime && delayFeedback && ...)
```

so `getParameter(g_paramReverbSend, 0)` returns `nullptr` and `getTailLengthSeconds()` silently stays `0.0` for the entire Osirus line — exactly the pre-PR behaviour. Only TI benefits.

**2. Part 0 is hardcoded.** `getParameter(g_paramReverbSend, 0)` always reads part 0, but `VirusProcessor::getPartCount()` supports up to 16 multitimbral parts and these parameters are per-part. In a Multi where part 0 has no reverb but another part does, the tail is under-reported and the host may cut the tail short on stop or during an offline bounce. Either take the max over active parts, or document the Single-mode-only assumption explicitly.

Smaller points for a resubmit:
- The comment says delay time is "0–1000 ms"; the actual TI value table tops out at **693.6 ms**.
- The new function is defined outside the `namespace virus { }` block using a qualified name, unlike everything else in that file.
- `kMaxReverbSeconds` / `kDelayTimeRangeMs` / `kMaxDelaySeconds` use a `k` prefix; this codebase uses plain camelCase for local `constexpr`.

**3. Packaging.** The branch is stacked, so the diff also carries the lunasvg/RmlUi submodule work, `build_mac.sh` and the midiClock changes from #266.

The underlying idea — reporting a real tail instead of a hardcoded `0` — is worth having. Please resubmit from current `main` with the classic-Virus path handled (or explicitly excluded) and the multi-part question resolved.

PR #268, opened 2026-06-01
Feature/CLAP : remote controls

kartun83:
## What this does

Implements `CLAP_EXT_REMOTE_CONTROLS` for OsTIrus, allowing CLAP hosts (Bitwig, Reaper, Carla etc.) to map hardware controllers to named groups of plugin parameters — no manual MIDI-learn required. Turning a knob on a
Push or similar surface immediately controls the parameter shown on the corresponding page.

## Pages (18 total, 8 slots each)

| Section | Pages |
|---|---|
| Oscillators | Oscillator 1, Oscillator 2 |
| Filter | Filter 1, Filter 2, Filter Envelope |
| Amplifier | Amplifier |
| LFO | LFO 1, LFO 2, LFO 3 |
| Effects | Chorus, Phaser, Reverb, Delay |
| FX1 | Filter Bank, Distortion, EQ |
| — | Arpeggiator, Performance |

Each oscillator and filter gets its own dedicated page so no knob slot is wasted on a mix of Osc1 and Osc2 controls. The FX1 section follows the hardware's Patch Distortion layout (curve, intensity, mix, treble booster, high cut, tone, bass intensity, punch). EQ covers all 7 hardware params (low/mid/high gain + frequency, mid Q) with one empty
slot. Filter Bank exposes the context-dependent params (frequency, filter type, slope) that change display name with the bank type but always map to the same hardware index.

Parameter names map exactly to entries in `parameterDescriptions_TI.json`.
Empty slots use `nullptr` (permitted by the CLAP spec).

## Dynamic part tracking

Pages always reflect the **currently selected part**, in both single and multi mode. When the user switches parts in the plugin UI, `onCurrentPartChanged` fires, `m_remoteControlsPart` updates, and `remoteControlsChanged()` tells the host to re-query all pages. The host then receives parameter pointers for the new part — so a hardware knob mapped to Cutoff always controls the part the user is editing.

Page IDs are encoded as `(pageIndex << 8) | part`, making them stable within a part while allowing the host to distinguish sets across part changes. This encoding also prepares the ground for a future `suggest_page` call that will hint the host to scroll to the relevant page when the user navigates to a section in the plugin UI.

## suggest_page on gesture

When the user touches any control in the plugin UI, `audioProcessorParameterChangeGestureBegin` fires. A reverse lookup map (built at startup from the static page table, zero runtime cost after init) finds the corresponding page and calls `suggestRemoteControlsPage()`. The host receives a hint to scroll to the relevant page automatically.

**Note:** `suggest_page` is a host-optional hint. Whether the host acts on it depends on its implementation. Hardware controller gestures are not routed back by the current clap-juce-extensions wrapper (`CLAP_EVENT_PARAM_GESTURE_BEGIN` has no case in `process_clap_event`), so controller-only users can navigate pages manually on their hardware.

## Settings toggle

A checkbox in Settings → GUI → "CLAP Controller Integration" enables
or disables the suggest_page behaviour. The preference is stored per
installation via `getConfig()` (not per project), defaulting to on.
Disabling it leaves pages fully functional for manual navigation.

## Architecture

`OsTIrusProcessor` (in `osTIrusJucePlugin`) inherits `clap_juce_audio_processor_capabilities` and overrides
`supportsRemoteControls()`, `remoteControlsPageCount()`, and `remoteControlsPageFill()`. The shared `virusJucePlugin` library remains CLAP-agnostic — no build system changes needed.

The page table lives in `VirusTIRemoteControls.h` as a `constexpr` array in the `virusTI` namespace. Adding or reordering pages in the future is a single-file edit with no processor code to touch.

`getControllerConst()` (added in the CLAP_EXT_TAIL PR) is used in `remoteControlsPageFill()` for const-correct parameter lookup without triggering lazy controller creation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
+
 # Enable RmlUi Lua bindings and point to our bundled Lua
 set(RMLUI_LUA_BINDINGS ON CACHE BOOL "" FORCE)
 set(RMLUI_LUA_BINDINGS_LIBRARY "lua_as_cxx" CACHE STRING "" FORCE)
+set(RMLUI_SVG_PLUGIN ON CACHE BOOL "" FORCE)
 
 # Create the Lua::Lua target so RmlUi's Dependencies.cmake finds it and skips find_package(Lua)
 add_library(Lua::Lua INTERFACE IMPORTED)
--- a/source/jucePluginLib/processor.h
+++ b/source/jucePluginLib/processor.h
@@ -89,6 +89,13 @@ namespace pluginLib
 			return m_controller.get();
 		}
 
+		// Const access to the controller without triggering lazy creation.
+		// Returns nullptr if the controller has not been created yet.
+		const Controller* getControllerConst() const
+		{
+			return m_controller.get();
+		}
+
 		virtual bool setLatencyBlocks(uint32_t _blocks);
 		virtual void updateLatencySamples();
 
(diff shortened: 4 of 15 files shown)


kartun83:
Related: dsp56300/clap-juce-extensions#1 makes hosts (e.g. Bitwig) actually follow preset loads on the CLAP param/remote-control stream. Without it, the remote-control pages added here map correctly but their knob positions don't update when a preset is loaded. The two changes complete each other.

dsp56300:
Closing on a build regression plus packaging — but the feature design here is good and we would like it back.

**Build break when CLAP is disabled.** `OsTIrusProcessor.h` and `VirusTIRemoteControls.h` both `#include "clap-juce-extensions/clap-juce-extensions.h"` unconditionally, and `OsTIrusProcessor` unconditionally inherits `clap_juce_extensions::clap_juce_audio_processor_capabilities`. But that header only exists in the build when CLAP is on — `source/CMakeLists.txt`:

```cmake
if(${CMAKE_PROJECT_NAME}_BUILD_JUCEPLUGIN_CLAP)
    add_subdirectory(clap-juce-extensions)
endif()
```

JUCE compiles the processor sources **once** into a shared target that every format links against, so building with `-Dgearmulator_BUILD_JUCEPLUGIN_CLAP=OFF` fails to compile the *entire OsTIrus plugin*, not just a CLAP target. That flag is documented and used, so this is a real regression. The fix is small: set a compile definition inside the existing `if(USE_CLAP)` block and guard the include, the inheritance and the new methods behind it.

**Two smaller things:**
- The new "CLAP Controller Integration" checkbox markup is added here, but its C++ wiring only exists in #269 — as submitted this PR ships an inert checkbox. Once wired, consider hiding the section for non-CLAP builds, the way the adjacent Animation section is conditionally hidden via `supportsLogoAnimation()`.
- `m_paramNameToPage.emplace(name, pageIdx)` keeps the *first* page for parameters that appear on several pages (e.g. `Patch Volume` / `Panorama` are on both Amplifier and Performance), so those will always suggest "Amplifier". Cosmetic.

**Packaging:** the branch is stacked, so the diff also carries lunasvg/RmlUi, `build_mac.sh` and the midiClock changes from #266.

Credit where due: the ~150-entry parameter table is accurate — we spot-checked awkward entries including the existing `Arp Pattern Selct` typo and the trailing space in `Dly Depth ` and they match the descriptors exactly. The threading is also right: `suggestRemoteControlsPage()` is main-thread-only by construction here, since gestures are only wrapped for `Origin::Ui`/`Unknown`. Please resubmit from current `main` with the `USE_CLAP` guard.

PR #269, opened 2026-06-02
Feature: panic skin buttons

kartun83:
### Expose panic actions to main skin via button IDs

- Moves panic logic (`panicAllNotesOff`, `panicNoteOffEveryNote`, `panicRebootDevice`) from `SettingsMidi` into `Processor` — single authoritative location
- `SettingsMidi` delegates to the new methods (no behaviour change for the settings page)
- Registers click handlers in `VirusEditor::create(` for `btPanicAllNotesOff`, `btPanicNoteOffEveryNote`, `btPanicReboot` — silently skipped when absent, so existing skins are unaffected

Skin usage — no params, no Lua needed:

```rml
<button id="btPanicAllNotesOff"      class="jucePos juceButton" isToggle="0" style="...">!</button>
<button id="btPanicNoteOffEveryNote" class="jucePos juceButton" isToggle="0" style="...">!</button>
<button id="btPanicReboot"           class="jucePos juceButton" isToggle="0" style="...">!</button>
```

### Why this matters

Stuck notes and runaway reverb/delay tails are a real problem during live performance and studio sessions. On the original Virus TI hardware, Access provided a dedicated panic shortcut (MONO+SYNC held simultaneously) for exactly this reason.

In the emulator the situation was worse than it might appear:

- **CC123 (All Notes Off)** from a hardware controller already works — it passes straight through to the firmware. But it does nothing to reverb/delay tails, which live in DSP memory and keep running regardless of voice state. Also, users without hardware controller have to dig their way through the settings for pulling killswitch.
- **The settings page panic buttons were the only way to truly stop sound** — requiring two or three clicks through a menu, far too slow when something goes wrong live.
- **Reboot Device** is the only option that actually silences everything — it recreates the DSP from scratch, equivalent to what the hardware **MONO+SYNC** shortcut likely does internally.
- This PR makes all three panic actions available as first-class skin buttons, so a skin developer can place an immediately accessible panic button directly on the instrument surface — the same way hardware synths expose it as a front-panel control. No new plugin parameters, no automation lanes, no MIDI learn slots consumed. Skins that don't use the buttons pay no cost.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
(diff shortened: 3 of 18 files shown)


dsp56300:
**The content of this PR is good and we want it.** Closing purely on packaging — please read to the end.

This is exactly the refactor direction we prefer: the MIDI panic logic previously lived inline in `SettingsMidi.cpp`, and you moved it into `pluginLib::Processor` so the settings page and the new skin buttons share one implementation instead of duplicating it. The `findChild("btPanic...", false)` wiring also matches the existing optional-element idiom used a few lines above for `VersionInfo` / `PresetSave`, and the buttons stay inert until a skin defines the IDs — a clean incremental rollout.

The blocker is that this branch is **stacked on your earlier branches** rather than on `main`. The diff re-contains the lunasvg/RmlUi submodule work, `build_mac.sh`, the midiClock changes from #266 and the `VirusProcessor`/`ParameterNames.h` changes from #267. That makes it unreviewable as a unit and is why it shows as CONFLICTING.

**Please resubmit rebased on current `main` with only the panic-related files** (`jucePluginLib/processor.cpp/.h`, `jucePluginEditorLib/settingsMidi.cpp`, `virusJucePlugin/VirusEditor.cpp`). We will merge it.

Minor nit for the resubmit: `panicRebootDevice()` is a one-line pass-through around the already-public `rebootDevice()` — keep it if you want the naming symmetry, but it is not required.

PR #270, opened 2026-06-03
Fix/callasync use after free

kartun83:
# Rapid preset switching: crash fixes, stability improvements, and handshake hardening

This PR addresses a cluster of issues that manifested as crashes, CPU spikes, audio deadlocks, and eventual unresponsiveness when switching presets rapidly in OsTIrus/Virus TI.

---

### Root crash: JIT out-of-bounds PC access

Using crash logs and lldb, the crash was traced to `dsp56k::DSP::execJit()` calling a **null function pointer** at address `0x0`.

The JIT function table (`m_jitFuncs`) is backed by a `std::vector` that grows on demand. When the DSP firmware jumps to a PC address beyond the current table size, `m_jitEntries[pc]` reads out-of-bounds, returning null (OS zero-maps freed pages), and the resulting call crashes at address `0x0`.

This happened because several code paths in `JitBlockChain::getChildBlock()` returned early without calling `ensureFuncSize()`, so branch targets outside the current table range were never registered.

### Secondary JIT issue: circular reference guard broken by pre-allocation

An earlier attempt to fix the OOB crash by pre-allocating `m_jitFuncs` to `pSize` accidentally broke the circular-reference guard in `getChildBlock()`.

The guard checked:

```cpp
m_jitFuncs[_pc] == nullptr
```

meaning "block is currently being generated and not yet compiled".

With pre-allocation, every entry starts as `&funcCreate`, never `nullptr`. This caused null function pointers to be **baked directly into parent JIT blocks' compiled code**, producing crashes at `0x0` through a different path.

### Recursive JIT compilation / stack overflow

The firmware contains self-modifying code patterns.

During preset loading, P-memory writes invalidate JIT blocks. If the firmware PC is at the same address being invalidated, `funcCreate` is called recursively. Without a guard, this grows the C stack unboundedly until a SIGSEGV occurs at the thread stack boundary.

### Firmware PC corruption → `0x400000` NOP sled

The deepest root cause: rapid preset switching corrupts the firmware's program counter (most likely through the race described below), sending it to `0x400000` — a region where P memory is all zeros.

The JIT compiles these as NOP blocks, DSP A stops producing audio output, DSP B blocks waiting for ESAI audio from DSP A, and both DSPs deadlock.

CPU spikes to 80–100% during compilation of the NOP sled, then drops as the cache fills.

### Microcontroller handshake race

In `sendPreset()`, `parser.waitForPreset(size)` was called **after** writing data to the HDI08 queue.

The DSP thread runs concurrently and can:

1. Consume the data
2. Process the preset
3. Send all confirmation bytes

before `waitForPreset()` is armed.

When `process()` checks:

```cpp
waitingForPresetReceiveConfirmation()
```

it sees `false` and sends the next preset mid-handshake.

The firmware then receives a new preset header while still reading the previous preset's data, corrupting its internal state machine.

### Queue accumulation

When rapidly clicking through presets:

```
A → B → C → D
```

all four were stored in `m_pendingPresetWrites` and sent sequentially.

The user only wanted D, but the DSP processed all four, creating unnecessary:

- P-memory writes
- JIT invalidations
- opportunities for handshake corruption

### Missing confirmation timeout

Once `m_remainingPresetBytes` becomes permanently stuck above zero (for example after DSP execution reaches `0x400000`), `waitingForPresetReceiveConfirmation()` returns `true` forever.

There was no recovery mechanism.

Subsequent preset changes accumulated in the queue and were never executed.

---

## What was implemented

| Layer | Change | File(s) |
|---------|---------|---------|
| JIT | `getChildBlock()` early-return paths now call `ensureFuncSize(_pc)` before returning `nullptr` | `jitblockchain.cpp` |
| JIT | Circular-reference guard changed from `m_jitFuncs[_pc] == nullptr` to `!e.block->getFunc()` | `jitblockchain.cpp` |
| JIT | Added `execOob()` which grows the JIT table and executes OOB PCs safely | `jit.cpp`, `jit.h`, `dsp.h` |
| JIT | Added recursion depth guard via `thread_local int s_depth` in `Jit::create()` | `jit.cpp` |
| JIT | `setJitEntries()` now takes `(data, size)` and stores `m_jitEntriesSize`; `execJit()` checks bounds and routes OOB PCs through `execOob()` | `dsp.h`, `jit.cpp` |
| DSP submodule | Reverted MmuArray pre-allocation because it capped growth at `pSize` and prevented access to firmware-used addresses such as `0x400000` | `mmuarray.h` |
| Microcontroller | Moved `waitForPreset()` before `writeRX()` so confirmation tracking is armed before data transmission | `microcontroller.cpp` |
| Microcontroller | Added single-mode queue coalescing: older pending single-mode presets are discarded and only the latest is kept | `microcontroller.cpp` |
| Microcontroller | Added confirmation timeout (`m_presetConfirmationTimeout`, default ~3 seconds) which resets parser state if confirmation never arrives | `microcontroller.cpp/.h`, `hdi08TxParser.h` |
| UI | Added configurable "Preset Confirmation Timeout" setting with Fast / Normal / Slow / Very Slow options | `tus_settings_dspaudio_OsTIrus.rml`, `SettingsDspAudioOsTIrus.cpp/.h`, `VirusProcessor.cpp/.h`, `device.cpp/.h` |

---

### Pre-allocating `m_jitFuncs` to the full DSP address space

Rejected because:

- macOS uses the vector-backed path rather than the MMU path
- 16 million entries would consume roughly 128 MB per DSP
- it incorrectly capped valid firmware addresses above `pSize`

### Redirecting OOB PC to `0x0` (reset vector)

Tested but rejected.

Restarting firmware from the reset handler broke audio initialization and produced silence.

### Killing the device on OOB zero-memory access

A `setDeviceInvalid()` path was explored but required cross-thread signaling that does not currently exist in the architecture.

### Fully synchronous preset handshake

The long-term ideal solution would require explicit DSP acknowledgment before transmitting the next preset.

This was not implemented because it would require significant HDI08 protocol redesign and could introduce audio-callback stalls or glitches.

---

### Firmware PC corruption remains unresolved

The DSP still occasionally jumps to `0x400000` during extremely aggressive preset switching.

The underlying issue appears to be a timing race in the HDI08 microcontroller ↔ DSP handshake.

Specifically:

```cpp
m_hdi08.exec()
```

runs without holding the mutex, allowing DSP state to change between execution and the subsequent guard checks.

A complete fix would require either:

- moving `exec()` inside the mutex (with deadlock risk), or
- redesigning the preset protocol around synchronous acknowledgments

Neither approach was considered safe for this branch.

### ESAI deadlock remains unresolved

When firmware reaches `0x400000`, DSP A executes NOPs and stops generating audio.

DSP B waits indefinitely for ESAI frames from DSP A, resulting in a deadlock.

The new timeout mechanism allows the microcontroller queue to recover, but it does not recover the DSP pair.

Audio must be restarted or the device rebooted.

### Non-determinism remains

Behavior still varies run-to-run because the remaining failure mode is timing-dependent.

Some runs switch dozens of presets successfully at moderate CPU usage, while others hit an unlucky scheduling window and trigger the corruption path.

The implemented changes significantly reduce the frequency of failures, but do not eliminate the underlying race condition.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
+cmake -G Xcode -S . -B ./temp/cmake_macos
+cmake --build ./temp/cmake_macos --config Release
+cd ./temp/cmake_macos
+CMAKE_BIN="$(dirname "$(command -v cmake)")"
+"${CMAKE_BIN}/cpack" -G ZIP
--- a/source/3rdparty/CMakeLists.txt
+++ b/source/3rdparty/CMakeLists.txt
@@ -18,9 +18,14 @@ add_library(Freetype::Freetype ALIAS freetype)
 # Lua 5.4 for RmlUi scripting
 add_subdirectory(lua)
 
+# lunasvg for RmlUi SVG plugin
+set(LUNASVG_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+add_subdirectory(lunasvg)
+
 # Enable RmlUi Lua bindings and point to our bundled Lua
 set(RMLUI_LUA_BINDINGS ON CACHE BOOL "" FORCE)
 set(RMLUI_LUA_BINDINGS_LIBRARY "lua_as_cxx" CACHE STRING "" FORCE)
+set(RMLUI_SVG_PLUGIN ON CACHE BOOL "" FORCE)
 
 # Create the Lua::Lua target so RmlUi's Dependencies.cmake finds it and skips find_package(Lua)
 add_library(Lua::Lua INTERFACE IMPORTED)
--- a/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
+++ b/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
@@ -226,9 +226,13 @@ namespace jucePluginEditorLib::patchManager
 			if(_part == -1)
 				return;
 
-			juce::MessageManager::callAsync([this, _part, _savedPatches]
+			auto* self = this;
+			juce::MessageManager::callAsync([self, _part, _savedPatches]
 			{
-				setSelectedPatch(_part, _savedPatches.front());
+				std::lock_guard<std::mutex> lock(getInstancesMutex());
+				if (getInstances().find(self) == getInstances().end())
+					return;
+				self->setSelectedPatch(_part, _savedPatches.front());
 			});
 		});
 	}
--- a/source/jucePluginEditorLib/patchmanager/patchmanager.h
+++ b/source/jucePluginEditorLib/patchmanager/patchmanager.h
@@ -116,15 +116,15 @@ namespace jucePluginEditorLib::patchManager
 
 		void setCurrentPart(uint32_t _part);
 
+		static std::mutex& getInstancesMutex();
+		static std::set<PatchManager*>& getInstances();
+
 	protected:
 		void updateStateAsync(uint32_t _part, const pluginLib::patchDB::PatchPtr& _patch);
 
 		void startLoaderThread(const juce::File& _migrateFromDir = {}) override;
 
 	private:
-		static std::mutex& getInstancesMutex();
-		static std::set<PatchManager*>& getInstances();
-
 		pluginLib::patchDB::SearchHandle getSearchHandle(const pluginLib::patchDB::DataSource& _ds, bool _selectTreeItem);
 
 		State m_state;
--- a/source/jucePluginEditorLib/patchmanagerUiRml/listitem.cpp
+++ b/source/jucePluginEditorLib/patchmanagerUiRml/listitem.cpp
@@ -250,11 +250,15 @@ namespace jucePluginEditorLib::patchManagerRml
 
 			const auto part = data->getPart();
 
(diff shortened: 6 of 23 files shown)


dsp56300:
**The callAsync fix in here is correct** — we traced the mutex ordering and it genuinely closes the race rather than narrowing it: the instance-set erase is the first statement of the destructor, and every guarded lambda holds the mutex across its whole body, so either the callback finds nothing or the destructor cannot proceed until the callback releases. The `listitem.cpp` approach of capturing the more stable `PatchManager*` and riding its existing guard, instead of inventing lifetime tracking for the transient UI element, is a good instinct.

Closing anyway, because the PR is not reviewable or mergeable as one unit: **8 commits, 26 files, at least six unrelated concerns** under a title that names one of them.

| Commit | Concern |
|---|---|
| `7495e2297`, `a7a431951` | the actual callAsync fix |
| `6a47ff8e2` | macOS build script change |
| `76c1eaca9` | new RmlUi SVG plugin + lunasvg submodule |
| `e942228e4` | MIDI clock phase/loop resync |
| `03208cfef` | CLAP tail length |
| `99c46c516` | dsp56300 submodule bump |
| `e3be3c3d8` | JIT OOB fixes + MC handshake hardening |

That last one is the most problematic to carry in under this title, since the PR description itself states it is incomplete — "Non-determinism remains", "ESAI deadlock remains unresolved", "Firmware PC corruption remains unresolved". We are not merging a partial fix for a different bug as a side effect of a use-after-free fix.

**Please resubmit the callAsync fix alone**, branched from current `main`, touching only `patchmanager.cpp/.h`, `patchmanagerUiRml/listitem.cpp` and `pluginEditor.cpp/.h`. Two changes we would want on that resubmit:

1. `Editor::getInstancesMutex()` / `getInstances()` are added as **public**, but every call site is inside `Editor`'s own members — make them private. (`PatchManager` genuinely needs public access for `listitem.cpp`; `Editor` does not.)
2. Both classes expose `getInstances()` as a **non-const `std::set<T*>&`**, so any caller can insert or clear entries in the registry that every guarded callback trusts. Prefer a `static bool isAlive(const T*)` accessor, or at least return by const reference.

Also worth folding in: the ~10-line mutex/set/ctor-insert/dtor-erase block is now copy-pasted between `PatchManager` and `Editor`, and the lock-and-check guard appears four times. A small shared helper would fit the project's no-duplication convention better than a second hand-written copy.

PR #271, opened 2026-06-04
Feature/clap juce extensions 1.2.7

kartun83:
## Summary

Updates the `clap-juce-extensions` submodule from its 1.2.0-era commit
to current `main` (18 commits). CLAP headers bumped 1.2.0 → 1.2.7.

## Changes in clap-juce-extensions

- Fix mis-handled STEPPED parameters (discrete selectors/modes)
- Fix parameter threading race condition
- Dismiss popup menus on guiDestroy
- clap_plugin_t::reset wired to AudioProcessor::reset
- Loop points set in AudioPlayHead::PositionInfo (arp sync benefit)
- Correct latency change protocol (request host restart)
- Linux FD support + POLLIN-only fix
- JUCE 8.0.11 compatibility
- Double precision audio processing support
- JUCE-style note name support
- Resolve CMake CMP0177 warning

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "source/3rdparty/freetype"]
 	path = source/3rdparty/freetype
 	url = https://github.com/freetype/freetype.git
+[submodule "source/3rdparty/lunasvg"]
+	path = source/3rdparty/lunasvg
+	url = https://github.com/sammycage/lunasvg.git
--- a/build_mac.sh
+++ b/build_mac.sh
@@ -1,4 +1,5 @@
-cmake -G Xcode -S . -B ./temp/cmake
-cd ./temp/cmake
-cmake --build . --config Release
-cpack -G ZIP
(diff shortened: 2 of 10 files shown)


dsp56300:
**No objection to the bump itself** — closing on packaging only.

We reviewed the 18 upstream commits between `2aaad9e` and `e1f67893c`: a parameter-threading fix, loop-point reporting in `PositionInfo`, a latency-restart hook, a voice-info-changed hook and JUCE 8.0.11 compatibility. Nothing in that range touches the extension surface we implement, so it is low risk and worth having.

The issue is that this branch is **stacked**, so the diff is not just the submodule pointer — it also re-carries the lunasvg/RmlUi work, `build_mac.sh`, the midiClock changes from #266 and the `VirusProcessor` changes from #267. A dependency bump should be a one-line diff that we can revert cleanly in isolation if it ever misbehaves.

**Please resubmit rebased on current `main` containing only the `source/clap-juce-extensions` pointer change.** We will take it after a CLAP smoke-test.

PR #273, opened 2026-07-03
docs: clarify MCP async write/read-back semantics and SysEx addressing

kartun83:
## Summary
- Document that `get_state`/`get_current_preset` (engine state) and `get_parameter`/`list_parameters`/`dump_all_parameters` (host-side parameter cache) are separate stores that update on different triggers, with a concrete polling recipe for verifying a `send_sysex` patch load.
- Document the `DUMP_SINGLE` byte layout (bank/program offsets, device ID, checksum range) for re-addressing a captured SysEx dump to the edit buffer, plus the `RequestSingle`/`RequestMulti` bytes needed to make a load visible to the parameter tools.
- Clarify that JSON-RPC `tools/call` results are returned in the `/message` POST response body, not pushed onto the `/sse` stream.

Symptom: `send_sysex` loads that didn't show up in `get_parameter`. Traced against the current `MidiRoutingMatrix`/`Controller`/`Microcontroller` source: the behavior is correct by design (mirrors real MIDI hardware not echoing back a dump you send it), so no engine/server code changed — this PR is documentation only.

## Test plan
- [x] Reviewed rendered Markdown for broken tables/anchors
- [ ] N/A — docs-only change, no build/runtime behavior affected


dsp56300:
Closing this one on scope rather than content.

The title describes a documentation change, but the PR touches **63 files** — of which only ~81 lines are actually `doc/mcp_server.md`. Bundled in are, among others, a new 460-line `clapPresetDiscovery.cpp` feature, CI workflow changes, `base.cmake`, virusLib microcontroller/romfile changes, the lunasvg submodule, and the accumulated content of several other open PRs.

A change set that size can't be reviewed, tested or reverted as a unit, and it can't be merged without dragging in work that belongs to separate PRs.

On the documentation itself, we would take **part** of it:

- **In scope:** the async write / read-back semantics of the MCP server — that `send_sysex` returning `{"success": true}` only means the bytes were forwarded, not that the device applied them, and that the read tools may not reflect the change yet. That behaviour is a property of the MCP server itself, applies to every synth, and is currently undocumented. A short docs-only PR covering that is welcome.
- **Out of scope:** the Virus/OsTIrus `DUMP_SINGLE` byte layout (bank/program/checksum re-addressing). The MCP server is generic and shared by all supported synths — Virus-specific SysEx formats do not belong in `mcp_server.md`. That material belongs with the Virus documentation, not in the generic MCP reference.

So: if you want to resubmit, please make it a docs-only PR against current `main` touching just `doc/mcp_server.md`, limited to the generic server behaviour. Anything else in here — the CLAP preset discovery feature in particular — should come as its own separate PR so it can be judged on its own merits.

PR #274, opened 2026-07-04
Bump freetype submodule: 2.13.3 -> 2.14.3

kartun83:
## Summary
- Bumps the vendored freetype submodule from 2.13.3+181 (May 2025) to 2.14.3+71 (~315 upstream commits), the current upstream master as of this PR.
- Picks up a number of integer/signed-overflow hardening fixes in the autofit, truetype, and psaux modules along with the regular 2.13 -> 2.14 feature work.
- No changes needed on the gearmulator side: the `FT_DISABLE_*`/`SKIP_INSTALL_*` cache variables and the `freetype` CMake target name used in `source/3rdparty/CMakeLists.txt` are unchanged upstream.

## Test plan
- [x] Reconfigured and built all synths (Osirus, OsTIrus, Vavra, Xenia, NodalRed2x, JE8086) across AU/VST3/CLAP on macOS -- clean build, no freetype-related warnings or errors.
- [x] `auval` validation passed for the Osirus and Vavra AU plugins (instantiation, render, MIDI, parameters).
- [x] Loaded Osirus and OsTIrus (CLAP) in Bitwig -- skin/text rendering, resize, and MIDI/audio all confirmed working, no regressions from the freetype bump.

### DIFF
--- a/doc/mcp_server.md
+++ b/doc/mcp_server.md
@@ -66,9 +66,11 @@ The server uses HTTP with Server-Sent Events (SSE):
 
 | Endpoint | Method | Description |
 |---|---|---|
-| `/sse` | GET | SSE stream for receiving server events |
-| `/message` | POST | Send JSON-RPC 2.0 requests |
+| `/sse` | GET | SSE stream. Only carries the initial `endpoint` event and periodic keepalives — **not** tool call results. |
+| `/message` | POST | Send JSON-RPC 2.0 requests. **The JSON-RPC response (including `tools/call` results) is returned directly in this POST's response body.** |
 | `/` | GET | Health check (returns server info) |
+
+> **Note for spec-compliant SSE clients:** unlike some MCP SSE transports, this server does not push `tools/call` results onto the `/sse` stream. Send the request to `/message` and read the result from the HTTP response of that same request.
 
 ### Protocol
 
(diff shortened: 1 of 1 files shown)


dsp56300:
Leaving this open, but it needs to be split before we can take it.

The title says "Bump freetype submodule 2.13.3 -> 2.14.3", and we have no objection to that — but the submodule pointer is 1 line of the diff, and the other ~100 lines are MCP documentation. Those are two unrelated changes: one is a dependency bump we might need to revert on its own if a font regression shows up, the other is prose. They should not share a PR.

Please split into:

1. **The freetype bump alone** — just the `source/3rdparty/freetype` pointer. We will take that after a build check.
2. **A docs-only PR** for the `doc/mcp_server.md` content, with one scoping change described below.

On the documentation, we want part of it but not all:

- **In scope:** the async write / read-back semantics — that `send_sysex` returning `{"success": true}` only confirms the bytes were forwarded, not that the device accepted or applied them, and that the read tools may not reflect the change yet. That is generic MCP server behaviour, true for every synth, and currently undocumented. Useful, please keep it.
- **Out of scope:** the Virus/OsTIrus `DUMP_SINGLE` byte layout (bank / program / checksum re-addressing). The MCP server is generic and shared by all supported synths, so synth-specific SysEx formats do not belong in `mcp_server.md`. That belongs with the Virus documentation instead.

Same note as on #273, which contains a superset of this content — whichever of the two you continue with, one docs-only PR against current `main` is enough.

PR #275, opened 2026-07-04
Fix Patch Manager search-handle leaks causing unbounded CPU growth

kartun83:
## Summary

Two related leaks in the Patch Manager's search-handle bookkeeping (`pluginLib::patchDB::DB`) that cause plugin host CPU usage to climb after loading multiple banks / switching many patches, and never recover:

- `PatchManager::updateStateAsync()` only called `cancelSearch()` on the "found a datasource match" path; the "no match found" early return skipped it, so the `Search` stayed in `DB::m_searches` forever. Since `DB::updateSearches()` rescans every entry in `m_searches` (including leaked ones) against every newly loaded patch on each subsequent bank load/patch change, leaked searches make every future load do strictly more work, with cost proportional to cumulative history rather than current state.
- `DB::cancelSearch()` always inserted the handle into `m_cancelledSearches`, but that set is only ever pruned inside `executeSearch()`'s per-datasource cancellation check, which never runs for a search that has already completed — the common case for essentially every Patch Manager UI interaction (tree/list rebuilds, filtering, patch selection). This is a genuine unbounded memory leak, distinct in character from the first (insert/find/erase only, never iterated, so not itself a CPU driver).

Both were found while investigating Bitwig host CPU climbing to 150-170% and never dropping back after loading several banks and switching patches in Osirus/OsTIrus.

## Changes

- `updateStateAsync()`: dispatch `cancelSearch(handle)` on the empty-results path too, mirroring the already-correct success path.
- `cancelSearch()`: only track a handle in `m_cancelledSearches` if the search hasn't completed yet, since a completed search will never be looked at again.

## Test plan

- [x] Builds clean (`jucePluginLib`, `jucePluginEditorLib` targets, Xcode/macOS)
- [x] Reproduced the original symptom (Osirus/OsTIrus CLAP in Bitwig: load several banks, switch many patches, CPU climbs and doesn't recover) and confirmed CPU stays flat after this fix

### DIFF
--- a/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
+++ b/source/jucePluginEditorLib/patchmanager/patchmanager.cpp
@@ -644,7 +644,16 @@ namespace jucePluginEditorLib::patchManager
 			results.assign(_search.results.begin(), _search.results.end());
 
 			if(results.empty())
+			{
+				// the search found no match, but it must still be cancelled/removed from the DB's
+				// search list, otherwise it leaks forever and every subsequent bank load / patch
+				// update has to scan it in DB::updateSearches, causing CPU usage to grow unbounded
+				runOnUiThread([this, handle]
+				{
+					cancelSearch(handle);
+				});
 				return;
+			}
 
 			if(results.size() > 1)
 			{
--- a/source/jucePluginLib/patchdb/db.cpp
+++ b/source/jucePluginLib/patchdb/db.cpp
@@ -480,7 +480,14 @@ namespace pluginLib::patchDB
 			return;
 
 		std::unique_lock lock(m_searchesMutex);
-		m_cancelledSearches.insert(_handle);
+
+		// only remember the handle as cancelled if the search may still be picked up by executeSearch
+		// later (i.e. it's still queued/running). A search that has already completed will never be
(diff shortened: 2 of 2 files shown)


dsp56300:
Merged — thank you.

This is exactly the kind of PR that is easy to say yes to: one concern, two files, branched cleanly off `main`, and the fix is right. We verified both halves against the current tree:

- `updateStateAsync()` really did return on `results.empty()` without calling `cancelSearch()`, so the handle stayed in `DB::m_searches` forever — only `cancelSearch()` ever erases from that map — and every later `updateSearches()` rescanned it.
- The second half is the more valuable one. Because `Search::setCompleted()` invokes its callback synchronously, the search is already `Completed` by the time `cancelSearch()` runs, and `findDatasourceForPatch` searches never go through `searchInDs()`'s cleanup — so *every* such call was leaking into `m_cancelledSearches` before your `state != SearchState::Completed` guard. Good catch; that one was not obvious.

Wrapping the new `cancelSearch()` in `runOnUiThread()` to match the adjacent branch was the right call too.

Thanks also for #277 and #278 — both are being handled separately (#278 has review notes on it). And apologies for the slow turnaround on your PRs generally; the ones that were closed were closed on packaging, not on the quality of the work, and several of them we would take straight away if resubmitted rebased on `main`.

PR #277, opened 2026-07-05
Fix MIDI-Learn listener leak and prevent macOS App Nap throttling

kartun83:
## Summary
- `MidiLearnTranslator::unsubscribeFromParameters()` removed listeners by indexing into the *current* mapping list, but `subscribeToParameters()` skips mappings whose parameter can't be resolved. Once the mapping list and listener-ID list went out of sync this way, any subsequently-skipped or reordered mapping caused its listener to never be removed - a permanent leak. Fixed by storing each listener together with the exact parameter it was attached to, so removal no longer depends on the mapping list's shape.
- Added a `juce::ScopedLowPowerModeDisabler` member to `Processor`, held for the plugin's lifetime, to opt the process out of macOS App Nap. Without it, macOS can demote the whole plugin-host process into a lower-power scheduling state once it decides the process isn't interactive, independent of any individual thread's own QoS/realtime priority settings.

Both were found while investigating a real-world CPU usage report on OsTIrus in Bitwig on macOS. Neither turned out to explain that specific symptom (which further investigation points to being macOS/SoC-level clock frequency scaling, tracked separately), but both are real, independent bugs/hardening worth fixing regardless.

## Test plan
- [x] Clean Release build on macOS (Xcode/arm64+x86_64 universal), `jucePluginLib` target
- [ ] CI build across all platforms/synths

### DIFF
--- a/source/jucePluginLib/midiLearnTranslator.cpp
+++ b/source/jucePluginLib/midiLearnTranslator.cpp
@@ -355,7 +355,7 @@ namespace pluginLib
 	void MidiLearnTranslator::subscribeToParameters()
 	{
 		const auto& mappings = m_preset.getMappings();
-		m_paramListenerIds.reserve(mappings.size());
+		m_paramListeners.reserve(mappings.size());
 
 		for (const auto& mapping : mappings)
 		{
@@ -380,23 +380,19 @@ namespace pluginLib
 					onParameterChanged(paramName, _param->getValue(), origin);
 				});
 
-			m_paramListenerIds.push_back(listenerId);
+			m_paramListeners.emplace_back(param, listenerId);
 		}
 	}
 
 	void MidiLearnTranslator::unsubscribeFromParameters()
 	{
-		const auto& mappings = m_preset.getMappings();
-		
-		// Remove listeners
-		for (size_t i = 0; i < m_paramListenerIds.size() && i < mappings.size(); ++i)
-		{
-			auto* param = m_controller.getParameter(mappings[i].paramName, 0);
-			if (param)
-				param->onValueChanged.removeListener(m_paramListenerIds[i]);
-		}
+		// Remove each listener from the exact parameter it was attached to. This is independent
+		// of the current mapping list, so it stays correct even if mappings were skipped during
+		// subscribe or the preset changed in between.
+		for (const auto& [param, listenerId] : m_paramListeners)
+			param->onValueChanged.removeListener(listenerId);
 
-		m_paramListenerIds.clear();
+		m_paramListeners.clear();
 	}
 
 	void MidiLearnTranslator::onParameterChanged(const std::string& _paramName, float _normalizedValue, Parameter::Origin _origin)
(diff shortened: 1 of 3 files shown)


dsp56300:
We want the MIDI-Learn half of this — it is the best-diagnosed fix in your current batch. One change requested before merging, described at the end; it should take you about a minute.

**On the MIDI-Learn fix: it is more serious than the title says, and you fixed it correctly.** `subscribeToParameters()` skips pushing an ID whenever `feedbackTargets == 0` or the parameter cannot be resolved, while `unsubscribeFromParameters()` zips `m_paramListenerIds[i]` against `mappings[i]` by raw index — so the two go out of alignment the first time anything is skipped. From then on `removeListener()` is called with the wrong ID (a silent no-op on that event), and the *correct* listener is never removed. It survives `~MidiLearnTranslator()`, so the next value change on that parameter fires a stale lambda holding `this` into freed memory. That is a use-after-free, not just a leak. Pairing each listener with the `Parameter*` it was registered on removes the index assumption entirely, and we checked that caching the raw pointer is safe here — `Controller::registerParams()` allocates each `Parameter` once at startup and nothing recreates them.

**The request: please drop the App Nap commit from this branch.**

Two reasons.

*Scope.* It is an unrelated concern in a PR whose title and value are about a listener bug. We are trying to keep changes independently reviewable and revertible.

*Substance.* We are not convinced it is needed, and would want a concrete case before taking it. App Nap is a per-process feature and `NSProcessInfo` activities are process-wide, but VST2/VST3/AU all load *inside the host's process* — so the host's own activity already covers every plugin it hosts, and ours would just nest on top without changing anything. For what it is worth, JUCE never uses `ScopedLowPowerModeDisabler` itself anywhere in the framework, including the standalone wrapper, and we have no bug report of background throttling or dropouts on macOS.

There is also a lifetime problem with where it sits. As a `Processor` member the activity is held for the entire life of every plugin instance, not just while audio is being processed — so eight instances mean eight nested activities, and a project left open and idle keeps the machine out of App Nap indefinitely. That defeats the feature exactly when it should engage, which is a real battery cost on laptops.

If you *do* have a concrete case, we are genuinely interested — please open it as its own PR with the details (which host, in-process or sandboxed, and ideally a repro). The scenario where we would expect this to matter is out-of-process hosting, where the DAW's activity does not cover us: AUv3, plugin-sandboxing hosts, the standalone build, or our own bridge server in `source/bridge`, which runs the DSP emulation in a separate process that no DAW will ever declare an activity for. Note that the current placement in `Processor` would not help that last case, which is arguably the one we actually own.

No need to close and reopen: drop that one commit and force-push this branch, and we will merge it as soon as CI is green.

dsp56300:
Heads-up: the source tree was restructured by ownership and manufacturer in `275285257`, now on `main`, so this PR conflicts. Unrelated to the change requested above — it affects every open PR touching `source/`.

| old | new |
|---|---|
| `source/jucePluginLib/midiLearnTranslator.cpp/.h` | `source/framework/juce/jucePluginLib/midiLearnTranslator.cpp/.h` |
| `source/jucePluginLib/processor.h` | `source/framework/juce/jucePluginLib/processor.h` |

Include paths are unaffected: there is no blanket `source/` include root any more, but each library exports its own parent as its include root, so `#include "jucePluginLib/..."` keeps working.

The ask from my previous comment is unchanged and still the only thing between this and a merge: drop the App Nap commit, leaving just the MIDI-Learn listener fix. Since you have to rebase anyway, dropping that commit during the rebase costs nothing extra. The listener fix itself is accepted — it is a genuine use-after-free and we want it.

PR #278, opened 2026-07-05
Fix orphaned Patch Manager search on ListModel destruction

kartun83:
## Summary
- `ListModel` had no destructor, so a search still in flight when the patch manager panel/document is torn down (e.g. closed mid-search) never called `cancelSearch()` - its handle stayed registered in `DB::m_searches` forever and the background search job kept running, the same class of bug as the already-fixed Patch Manager search leak, just on a different teardown path.
- Adds `~ListModel()` that calls the existing `cancelSearch()`.

## Test plan
- [x] Added `patchDbTest`, a standalone console test (mirrors `midiLearnTest`) verifying: `DB::cancelSearch()` removes a handle from `DB::m_searches`; an uncancelled handle stays registered indefinitely (demonstrating the leak this guards against); a `ListModel`-like RAII owner cancels its search on destruction.
- [x] `patchDbTest` built and run locally, all assertions pass.
- [x] `jucePluginEditorLib` builds clean with the destructor added.

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -78,6 +78,7 @@ if(${CMAKE_PROJECT_NAME}_BUILD_JUCEPLUGIN)
 	add_subdirectory(jucePluginData EXCLUDE_FROM_ALL)
 	add_subdirectory(pluginTester)
 	add_subdirectory(midiLearnTest)
+	add_subdirectory(patchDbTest)
 	include(juce.cmake)
 endif()
 
--- a/source/jucePluginEditorLib/patchmanagerUiRml/listmodel.cpp
+++ b/source/jucePluginEditorLib/patchmanagerUiRml/listmodel.cpp
@@ -49,6 +49,14 @@ namespace jucePluginEditorLib::patchManagerRml
 		});
 	}
 
+	ListModel::~ListModel()
+	{
+		// setContent()/cancelSearch() already cover replacing the current search, but nothing
+		// previously ran on teardown, so a search still in flight when the panel is destroyed
+		// (e.g. closed mid-search) kept running and its handle stayed in DB::m_searches forever.
+		cancelSearch();
+	}
+
 	PatchManagerUiRml& ListModel::getPatchManager() const
 	{
 		return m_patchManager;
--- a/source/jucePluginEditorLib/patchmanagerUiRml/listmodel.h
+++ b/source/jucePluginEditorLib/patchmanagerUiRml/listmodel.h
@@ -50,6 +50,7 @@ namespace jucePluginEditorLib::patchManagerRml
 		using Patches = std::vector<Patch>;
 
 		explicit ListModel(PatchManagerUiRml& _pm, juceRmlUi::ElemList* _list);
+		~ListModel();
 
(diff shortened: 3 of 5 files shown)


dsp56300:
Keeping this open — the production fix is correct, but the test that ships with it cannot pass in a Debug build.

**The fix is good.** `ListModel` has no destructor today, so a search started by `setContent()` that is still in flight when the panel is torn down never gets cancelled. `~ListModel() { cancelSearch(); }` reuses the existing no-arg member, runs on the message thread like the rest of the class, and complements #275 (same leak class, different trigger: UI teardown vs. empty result set) without overlapping it.

**The blocker: `patchDbTest` aborts on the first test in Debug.** None of the three tests call `db.startLoaderThread()`, but every `TestDb` destructor calls `stopLoaderThread()`, which relies on `JobQueue::destroy()` setting `m_destroy` — and that flag is only ever set by a queued closure that a live worker thread has to dequeue:

```cpp
// jobqueue.cpp
[this]{ m_destroy = true; }
```

With no thread started, `m_threads` is empty, the `join()` loop is a no-op, and `m_funcs.clear()` discards that closure unexecuted, so `m_destroy` stays `false`. `DB::~DB()` then hits:

```cpp
assert(m_loader.destroyed() && "stopLoaderThread() needs to be called by derived class in destructor");
```

which `abort()`s the process — it is not an exception, so `main()`'s try/catch does not help. That happens at the end of test 1, so tests 2 and 3 never run. It is masked in Release only because the assert compiles out, but `ctest -C Debug` is a normal thing to run here. Fix should just be calling `db.startLoaderThread();` after constructing each `TestDb`.

**Smaller point:** `testUncancelledSearchStaysRegisteredForever()` asserts `getSearch(handle) != nullptr` twice in a row with nothing happening in between, so it does not actually demonstrate persistence — compare with its two sibling tests, which do.

The `patchDbTest/CMakeLists.txt` scaffold itself is a clean clone of the existing `midiLearnTest` setup — no complaints there. Fix the loader-thread start and this is mergeable.

dsp56300:
Heads-up: the source tree was restructured by ownership and manufacturer in `275285257`, now on `main`, so this PR conflicts. Unrelated to the review feedback above — it affects every open PR touching `source/`.

| old | new |
|---|---|
| `source/jucePluginEditorLib/patchmanagerUiRml/listmodel.cpp/.h` | `source/framework/juce/jucePluginEditorLib/patchmanagerUiRml/…` |
| `source/patchDbTest/` (new dir) | `source/framework/juce/patchDbTest/` would match its dependency, or place it beside the other framework tools |

`source/CMakeLists.txt` is still where targets are registered, but the paths in it are now prefixed — for reference, the framework's own unit test is registered as `framework/tools/midiLearnTest`, which is the closest sibling to what you are adding. Include paths are unaffected, since each library exports its own parent as its include root.

The blocker from my previous comment still stands: `patchDbTest` never calls `db.startLoaderThread()`, so `DB::~DB()`'s assert aborts the process at the end of the first test in a Debug build and tests 2 and 3 never run. The `~ListModel()` fix itself is correct and we want it. Rebase, add the `startLoaderThread()` call, and it is mergeable.

PR #281, opened 2026-08-01
virusLib: name ControlCommand 76 as PURE_TUNING

insolentconstruct:
`ControlCommand` index 76 (`0x4C`) on the Global Master page is the Pure Tuning control. Two
independent sources agree on the name:

- **Access's own parameter database**, shipped inside the Virus Control installer, names this
  index `Global Master/Pure Tuning` and assigns it the value list `Val Num PureTuning`.
- **A sweep of a real Virus TI2 over MIDI** produced 285 live-edit messages on this index with
  116 distinct values, every one inside the declared range, with the front panel showing
  "Pure Tuning" throughout.

One detail worth having in the enum's vicinity for whoever reads it next: this is a **continuous
blend, not a mode switch**. `0x00` displays as *Tempered*, `0x40` as *Natural* and `0x7F` as
*Pure*, with every intermediate step showing its own number.

No behaviour change — the enumerator's value is untouched, only its name.


dsp56300:
Merged — thank you.

We confirmed the name against the codebase's own evidence rather than taking it on trust. `source/virusLib/microcontroller.cpp` contains the original TI boot-init SysEx stream, and one of its entries is:

```
0xF00020, 0x330110, 0x73404C, 0x00F700,   // Pure Tuning = 0
```

`0x4C` is 76 — exactly the value of `ControlCommand::UNK76`. That `0x73` + hex-offset addressing is reliable elsewhere in the same table (`0x73406A` → 106 `MIDI_CLOCK_RX`, `0x734079` → 121 `PANEL_DESTINATION`, `0x73407F` → 127 `MASTER_VOLUME`, `0x73405C` → 92 `MASTER_TUNE`), so it is a solid Rosetta stone. The label was sitting in that comment the whole time and simply never made it into the enum.

Nice piece of archaeology — small, verifiable, and the kind of change that makes the next person's job easier. `#282` is merged as well.

PR #282, opened 2026-08-01
virusLib: name ControlCommand 109 as DSP_CLOCK_ADJUSTMENT

insolentconstruct:
`microcontroller.cpp`'s `sendInitControlCommands` boot array already carries a comment against
this exact page and index (`0x734x`, 109) describing it as a DSP clock adjustment in percent,
default `0x64` = 100%, maximum `0x79` = 121%. Only the enumerator was still unattributed, so the
name and the neighbouring comment disagreed with each other.

Rather than take the comment's word for it, the command was exercised against the emulated DSP.
Sent via `PARAM_CHANGE_D`, it reprograms the PLL: values 100, 110 and 121 produce PLL `MF`
register values `0x8e`, `0x98` and `0xa3` (133.594, 143.002 and 153.35 MHz) — an exact
`mf = value + 42` relationship, tighter than the comment's "roughly".

As a control, sending the same value to the neighbouring unnamed index 108 produced the same
generic clock-recompute log line but left the clock at its default — so the effect belongs to
index 109 specifically, and is not a side effect of any write to that page.

No behaviour change — the enumerator's value is untouched, only its name.


dsp56300:
Merged — thank you.

Same verification as `#281`, and this one is even more direct. `source/virusLib/microcontroller.cpp`:

```
0xF00020, 0x330110, 0x73406D, 0x64F764,   // DSP clock adjustment in percent (roughly), default $64 = 100%, max $79 = 121%
```

`0x6D` is 109, exactly `ControlCommand::UNK6d` — whose placeholder name literally encoded that same hex byte. Whoever added it derived the name from the offset but never carried the adjacent comment's label into the enum. Dropping the "in percent" unit from the name is fine and consistent with siblings like `MASTER_TUNE` and `INPUT_BOOST`, which also omit units.

Both symbols were unreferenced anywhere else in the tree, so these were risk-free renames.

Thanks for both of these. `#283` is a more substantial change and we are still working through it — it looks sound and does not affect ROM loading for users today, so no concerns so far; we will follow up on that PR separately.

PR #283, opened 2026-08-18
Virus: fix ROM hash calculation and fail closed on unknown hashes

insolentconstruct:
`ROMFile::getHash()` returned a default all-zero MD5 because its writer was lost when ROM loading was refactored to pass bytes directly into the constructor. As a result, `DspMemoryPatchSet::allowedTargets` could not reliably distinguish ROM content.

This change:

- computes the digest from the supplied ROM bytes before initialization can discard invalid data;
- rejects all-zero incoming hashes and all-zero allowed-target entries as unknown;
- adds a regression executable covering known digests, positive and negative target matches, empty ROMs, and both zero-hash directions.

## What the test covers

`m_romDataHash` was declared in `romfile.h`, returned by `getHash()`, and assigned nowhere in the tree, so every ROM reported the default-constructed MD5 — all zeroes — and the only consumer could not tell one ROM from another. The check ran, returned an answer, and every caller believed it had chosen.

## Why the rejection cases carry the weight

The only consumer is the `allowedTargets` gate, so before this change the gate could not fail. A patch set authored against a hash read out of a running build would carry the all-zero digest and then match every ROM ever loaded — the dangerous direction, and the one an author reaches by measurement rather than by computing the digest externally.

That is also why the test checks rejection and not only acceptance: "the right ROM is accepted" would still pass if the gate accepted everything, which is exactly the state being fixed.

## Evidence that the test can fail

Each of the two fixes was removed independently and the suite re-run, to check that the checks are not measuring one thing under thirteen names:

| removed | result | failing checks |
|---|---|---|
| the `m_romDataHash` assignment in `romfile.cpp` (the pre-fix state) | 7/13 | 6 |
| the zero-MD5 guard in `dspMemoryPatch.cpp` | 11/13 | 2 |
| neither | 13/13 | — |

The two sets are disjoint. Four checks are failed by neither, and they are labelled as such in the file rather than counted as coverage: determinism (both hashes were zero before the fix, and zero equals zero), the empty-ROM digest, the negative target match (which passed only because `g_patches` is empty upstream), and the all-zero-entry case.

## Validation

- Release build against Gearmulator main `803fc747` and DSP submodule `c051afad`
- `virusRomHashTests` passed 13/13
- re-verified after both rounds of review changes: Release build on Ubuntu 24.04 / GCC 13.3, `virusRomHashTest` 13/13, exit 0. No test logic was touched by either round — comments only, plus three locals renamed.
- `add_subdirectory(virusRomHashTest)` is unconditional, matching the sibling test targets in that file. Confirmed with a `-DBUILD_TESTING=OFF` configure: the target is generated there exactly as `virusTestConsole` and `virusIntegrationTest` are.

Performance: ROM hashing occurs during ROM loading and patch matching occurs during DSP boot, not per audio callback or emulated instruction. No locks, logging, allocation, or atomics were added to the steady-state audio path.


dsp56300:
The fix is correct and we want it, and the test can stay. One change requested: please cut the prose down hard.

**On the fix.** Confirmed both defects are live: `m_romDataHash` is declared and returned but assigned nowhere in the tree, so `getHash()` hands back a default-constructed MD5 — all zeroes, identical for every ROM — and `git log -S"m_romDataHash"` shows it has never had a writer since the file was introduced. Hashing `m_romFileData` before `initialize()` (which clears it on a parse failure) is the right call, and the `if(!m_romFileData.empty())` guard is load-bearing rather than decorative, since MD5 of an empty vector is a non-zero constant that would have defeated the new sentinel. We also verified this changes nothing for users today: `g_patches` is empty, the `applyDspMemoryPatches` call site in `device.cpp` is commented out, and `getHash()` has no other consumer. ROM acceptance is purely structural and untouched. Fixing the gate before it is wired up is the right order.

**On the test: keep it.** It is better built than most of what we have. The reference digests are computed outside the tree rather than self-referentially; `dataB` being `dataA` reversed means a "hash" that merely summed or counted bytes would still have to distinguish them; and both directions are covered with the DSP memory checked rather than just the return value. We particularly appreciated that arms 4 and 9 are explicitly labelled as passing in the broken state too, instead of being quietly counted as coverage. That is unusually honest and we would like it to stay that way.

**The request: comments.** 69 of the 206 lines in `romHashTest.cpp` are comments, and `romfile.cpp` gets a 21-line block for a 2-line change. This codebase's house style is terse, mostly single trailing lines (`CLOCK_TEMPO = 16, // this is actually in page B`), and this is a long way from it. The problem is not that the reasoning is wrong — it is that most of it is commentary about testing methodology rather than about the code, and it is source that everyone reads from now on.

Please keep these, they are genuinely non-obvious and would otherwise cost the next reader time:
- `static const` rather than `static constexpr`, because MD5's default constructor assigns in the body
- the "arrays, not pointers" note about MD5's `char const(&)[33]` constructor
- the `python3 hashlib` provenance for the two reference digests
- why the hash is taken before `initialize()`

Please move these into the PR description, where they make the case for the change rather than living in the tree forever:
- the ablation results table (STUB 1 → 7/13, STUB 2 → 11/13, …)
- "a check that cannot fail is worse than no check", "measuring one thing under thirteen names"
- the surrounding methodology prose in the file header

One vocabulary note: "reddens", "ablation" and "arms" are not used anywhere else in this repository, so a reader has to pick up a private terminology to follow the file. Plain wording ("this check fails if the hash assignment is removed") would carry the same information.

Trimming `romfile.cpp` to roughly three lines and cutting the test header to a short paragraph would be about right. Push that and we will merge.

Two small things while you are in there, neither blocking: `g_addr`/`g_old`/`g_new` in `main()` are plain locals using the `g_` prefix this codebase reserves for static/namespace scope; and `if(BUILD_TESTING)` around the `add_subdirectory` is not what the sibling test targets in that file do — they are added unconditionally.

insolentconstruct:
Thank you for this — it is the most useful review I have had on a patch, and the part I appreciated most is that you went and checked the defects independently rather than taking the description's word for it. The `git log -S"m_romDataHash"` confirmation and the note that the empty-vector guard is load-bearing because MD5 of an empty vector is a non-zero constant are both things I would rather have had confirmed by someone else than have believed on my own.

Pushed in e49d10de.

**Trimmed.** `romfile.cpp` is down from 21 comment lines to 3, keeping only why the hash is taken before `initialize()`. `romHashTest.cpp` comments went from 69/206 to 40/176; the header is a short paragraph. The ablation table, "a check that cannot fail is worse than no check" and "measuring one thing under thirteen names" have moved into the PR description.

**Kept, per your list**: `static const` vs `static constexpr`, the `char const(&)[33]` binding note, the `python3 hashlib` provenance for both digests, and why the hash precedes `initialize()`.

**Vocabulary**: "reddens", "ablation" and "arms" are gone. You were right that they are private terminology — they come from tooling in another project of mine and I had stopped seeing them. The labels on the checks that also pass in the broken state survive in plain wording, since you asked for those to stay.

**Both small things**: `g_addr`/`g_old`/`g_new` renamed — they were plain locals wearing a prefix this codebase reserves for static scope, and you are right that it reads wrong. `add_subdirectory(virusRomHashTest)` is now unconditional, matching `virusTestConsole` and `virusIntegrationTest` in that file.

One thing that needs flagging rather than burying: **removing `if(BUILD_TESTING)` invalidated a claim in my own PR description**, which said a `BUILD_TESTING=OFF` configure contains no `virusRomHashTest` target. That is no longer true and was the whole point of the change, so the description now says the opposite and records the check. I mention it because the old wording would otherwise have sat there reading as verified.

Re-verified after the changes: Release build on Ubuntu 24.04 / GCC 13.3, `virusRomHashTest` 13/13, exit 0, and a separate `-DBUILD_TESTING=OFF` configure generates the target exactly as the two sibling test targets do.

insolentconstruct:
Pushed a second cleanup pass in `04849ed7`, trimming the comments further. No test logic changed, only comments and three local renames, and the test still passes 13/13.

dsp56300:
Both passes look right, and the content is accepted — thank you. Nothing further is wanted on substance.

Unfortunately the timing is bad through no fault of yours: the source tree was restructured by ownership and manufacturer in `275285257`, which landed on `main` after your last push. Everything under `source/` moved, so this PR now conflicts. The change needed is purely relocation — no code or comment changes.

New homes for the files this PR touches:

| old | new |
|---|---|
| `source/virusLib/romfile.cpp` | `source/axel/virusLib/romfile.cpp` |
| `source/virusLib/dspMemoryPatch.cpp` | `source/axel/virusLib/dspMemoryPatch.cpp` |
| `source/virusRomHashTest/` | `source/axel/virusRomHashTest/` |

`source/CMakeLists.txt` is still where test targets are registered — there is no per-manufacturer CMakeLists. Your `add_subdirectory` line should sit with its siblings, which now read:

```cmake
add_subdirectory(axel/virusTestConsole)
add_subdirectory(axel/virusIntegrationTest)
```

so `add_subdirectory(axel/virusRomHashTest)` alongside those, still unconditional as you changed it.

One include to check while you are moving: the tree no longer has a blanket `source/` include root, but each library exports its own parent as its include root, so `#include "virusLib/romfile.h"` and `#include "baseLib/md5.h"` continue to work unchanged. Depth-relative includes such as `../../` would not, but your test does not use any.

Rebase onto current `main`, move the three paths, and we will merge. Apologies for making you redo the packaging after two clean-up passes — the review feedback itself still stands as accepted.

Noting for the record, since you raised it yourself: flagging that removing `if(BUILD_TESTING)` falsified a claim in your own PR description, rather than leaving the old wording sitting there reading as verified, is the right instinct and was noticed.

dsp56300:
Merged — thank you, and sorry for making you rebase this after two clean-up passes. The tree move was our timing, not anything you did.

Final state is what we asked for: `romfile.cpp` down to three comment lines keeping only why the hash is taken before `initialize()`, the test header cut to a short paragraph, the private vocabulary gone, the methodology moved into the PR description where it makes the case without living in the tree forever, and the four technical notes we asked you to keep still there.

What we are actually taking, restated so it is on the record:

- `ROMFile::getHash()` had no writer anywhere in the tree, so every ROM reported the default-constructed all-zero MD5. `git log -S"m_romDataHash"` shows it never had one since the file was introduced, so this was latent rather than a regression.
- The empty-vector guard is load-bearing, not decorative — MD5 of an empty vector is a non-zero constant, which would have defeated the new sentinel for `ROMFile::invalid()`.
- The gate now refuses the all-zero digest in both positions, so a patch set authored by reading `getHash()` out of a broken build can no longer match every ROM ever loaded. That failure mode — silent DSP memory writes against the wrong firmware — is exactly the one worth closing before the feature is wired up rather than after.

No user-visible change today, as we verified earlier: `g_patches` is empty, the `applyDspMemoryPatches` call site in `device.cpp` is commented out, and ROM acceptance is structural and untouched. That is the right time to fix a gate.

Two things we appreciated in how you handled the review, independent of the code. You flagged that removing `if(BUILD_TESTING)` falsified a claim in your own PR description and corrected it, rather than leaving wording that would have read as verified to the next person. And when we asked for the comments to be cut, you measured the house style instead of estimating it. Both are habits we would rather encourage than take for granted.

PR #284, opened 2026-08-28
Feature: Add the ability for skins to hook into a per-frame DOM event

eynsai:
**Motivation**

Skins can animate declaratively with RCSS `transition` and `@keyframes`, but there is no way to run script on a frame boundary. Anything whose content must be computed each frame — an oscilloscope, a meter with its own ballistics — has no supported mechanism.

**Change**

`RmlComponent` dispatches a frame event to the document before each context update, carrying the elapsed seconds since the previous frame as `dt`:

```
<body onframe="onFrame(event) ...">

function onFrame(event)
  elapsedTime = elapsedTime + event.parameters['dt']
  scopeCanvas:repaint()
end
```

It is dispatched right after `evPreUpdate`, so a `repaint()` from the handler is drawn in the same frame.

**Opt-in and backwards compatibility**

The `onframe` attribute is the opt-in. A document without one gets no dispatch, no `enqueueUpdate()`, and the renderer keeps idling at 2 Hz exactly as before — existing skins are unaffected and pay nothing.

**Consistency and safety**

Consumed exactly like `onclick` / `onshow` / `ontransitionend`, so it needs no new concepts and no new global. Inline handlers are compiled into Lua event listeners invoked through `Interpreter::ExecuteCall`, so the existing 3-second `LUA_MASKCOUNT` script timeout and traceback already cover a runaway frame handler — no special handling added.

**Notes**

- The event is dispatched on the document and does not bubble, so the handler must be on `<body>`. Attaching one via `AddEventListener("frame", ...)` alone will not enable it, as the attribute gates dispatch.
- Frame intervals are not uniform; handlers should advance animation by dt rather than assume a fixed step. Documented.

**AI Disclosure**

This PR was made with the help of an AI coding assistant.

### DIFF
--- a/doc/lua_scripting.md
+++ b/doc/lua_scripting.md
@@ -31,6 +31,7 @@ Add a `<script>` block in your RML file's `<head>` section:
 - Use `onload="myFunction()"` on `<body>` to run code after all elements are created but before layout is computed.
 - Use `onshow="myFunction()"` on an element to run code when that element becomes visible. Layout is computed at this point, so `offset_width`/`offset_height` return correct values.
 - Inline event handlers like `onclick="..."` receive `event`, `element`, and `document` as parameters.
+- Use `onframe="myFunction(event)"` on `<body>` to run code before every rendered frame, after the document is loaded and laid out. See [Per-Frame Updates](#per-frame-updates-onframe).
 
 ### Globals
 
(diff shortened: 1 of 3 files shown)


eynsai:
Closing due to b0fe0cb.

PR #286, opened 2026-08-30
Add per-parameter MIDI input suppression

ChasonDeshotel:
This is a result of repeated questions in Discord about MIDI controls falling through to their default mappings.

We can't alter the firmware; however, if we simply refuse to forward those MIDI events, I believe we can satisfy the actual user need.

This adds a per-parameter `Disable MIDI Control` option to the existing context menu and uses the existing MIDI mapping machinery to discard matching input.

Once toggled, MIDI events are discarded, and the context menu label is changed to `Enable MIDI Control`.

An entry is also added to the `Mappings` table in Settings. Unchecking `Disabled` for a parameter that does not have a learned MIDI control will remove the entry from the table, since it no longer has a mapping or suppression state to display.

Style changes were also made to the `Mappings` table to prevent the `Remove` button from running off the page.

**Review note**: MIDI mapping type serialization
(source/framework/juce/jucePluginLib/midiLearnMapping.cpp:117-123)

This also changes MIDI mapping type serialization so that `Type::Invalid` is represented as `Invalid`, rather than falling back to `ControlChange`. This is used for suppression-only entries that do not have a learned MIDI control. Unknown types now also fall back to `Invalid`.

On the surface, this appears to be more correct and an improvement, but it is unknown if any legacy code depends on the existing behavior. This specific change deserves careful consideration during review.

Tested on Apple Silicon, macOS 15.7.4, Live 12.4.3

### DIFF
--- a/source/framework/juce/jucePluginData/tus_settings_midilearn.rcss
+++ b/source/framework/juce/jucePluginData/tus_settings_midilearn.rcss
@@ -28,11 +28,15 @@
 	padding-bottom: 0.3em;
 }
 
-.ml-col-type { width: 7%; }
+#mappingTable {
+	column-gap: 0.65em;
+}
+
 .ml-col-channel { width: 8%; }
 .ml-col-part { width: 8%; }
 .ml-col-controller { width: 9%; }
 .ml-col-mode { width: 17%; }
 .ml-col-invert { width: 7%; }
-.ml-col-parameter { width: 19%; }
-.ml-col-remove { width: 15%; }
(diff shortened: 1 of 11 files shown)


ChasonDeshotel:
I've been trying to understand the use case around this more. The Discord posts I used to draft the idea have since been deleted. The core of the idea is essentially to act as a MIDI filter and selectively discard events. It might be better as a global toggle that does something like discard all MIDI events that don't have a custom mapping.

I'm more than happy to revise for the smoothest user experience

PR #291, opened 2026-08-31
Refactor: Unify Osirus/Ostirus onto the shared controller mapping system

eynsai:
## Background

Every synth ships a `parameterDescriptions_*.json`. Alongside the parameter list it has a **`controllerMap`** — a plain table mapping CC values to synth-specific parameters. When MIDI arrives, the shared `pluginLib::Controller::parseControllerMessage` consults that table and updates the matching parameter, which is what makes knobs move in the UI.

Osirus and Ostirus don't do that. They have their own `parseControllerDump`, which skips the table and works *positionally*: CC number *n* is simply parameter *n* on page A, and poly-pressure note *n* is parameter *n* on page B.

Both of them nonetheless ship parameter JSONs containing an unused `controllerMap`.

This change was motivated as preparation for later PRs aimed at extending the controller to expose additional MIDI information to skins for visualization purposes. However, it also interacts with a couple of separately-submitted bugfix PRs.

## Why the switch is safe

The obvious worry is that the table and the positional code might disagree somewhere, and that flipping to the table would silently change which parameter a
controller drives. However, comparing every `controllerMap` entry against what the positional code actually does reveals that all 211 CC entries and all 197 poly-pressure entries agree between the positional and controller maps across both models.

Although there are no explicit disagreements, the JSON controller maps *are* missing a handful of parameters. This is addressed in the first commit in this PR.

### 1: Complete the ABC/TI controller maps

The maps are currently missing some entries covered by the positional path; this commit adds them. See the "Map delta" section below for full details.

### 2: Route MIDI input through the shared path

`parseControllerDump` is deleted, because the shared path already correctly handled what it decoded, making this plugin-specific special casing unnecessary.

Instead, `getPartsForMidiEvent` becomes virtual, providing a standardized interface for individual plugins to define how they address parts. `virus::Controller` overrides it because part addressing genuinely differs by message type: program change and bank select resolve through the **Part MIDI Channel** setting, page A and B parameter traffic in multi mode instead treats the raw channel nibble as a **part index**, and in single mode the device rejects anything that does not arrive on the **global MIDI channel**.

Program change and the poly pressure source check stay in the override rather than moving into the map, since neither is expressible as a parameter mapping - a program change triggers an action rather than a parameter write, and poly pressure (which actually carries dump data) has to be gated on where the message came from, which the map has no notion of.

## Map delta

This section details the full changes to the controller maps.

### Entries added

17 on Osirus, 20 on OsTIrus. Every one of them restores a controller the deleted
positional path already handled, so none of this is new reach — it is the same
coverage, written down declaratively.

| entry | Osirus C | OsTIrus TI | note |
|---|---|---|---|
| `cc 1` Modulation Wheel | added | added | performance controller |
| `cc 2` Breath Controller | added | added | performance controller |
| `cc 3` Contr 3 | added | added | performance controller |
| `cc 4` Foot Controller | added | added | performance controller |
| `cc 5` Portamento Time | already mapped | added | performance controller |
| `cc 122` Keyb Local | added | not mapped, see below | local control state |
| `pp 51`–`53` Soft Knob 1–3 ShortName | no such parameters on ABC | added | soft knob labels |
| `pp 112`–`121` SingleName0–9 | added | added | patch name characters |
| `pp 123` Category1 | added | added | patch category |
| `pp 124` Category2 | added | added | patch category |

`cc 122` is mapped on Osirus only. TI does have a `Keyb Local` parameter, but it sits on page D rather than page A, so TI has no page-A slot 122 for the map to address. The positional path found nothing there either, and testing confirmed `cc 122` is inert on TI both before and after this change.

Every parameter in the table above is a byte of the `singledump` packet, so the device stores them in the patch. That has a visible consequence which predates this PR: moving the mod wheel writes `Modulation Wheel`, singledump byte 1, so playing the wheel modifies the loaded patch and a later save stores its position. The positional path did the same — `findSynthParam(channel, PAGE_A, 1)` resolves to that parameter — so this PR neither introduces nor changes it.

### Osirus: page A slots left uncovered

| cc | parameter | device accepts it | reserved MIDI meaning | left alone because |
|---|---|---|---|---|
| 0 | Version | no | Bank Select MSB | reserved message would write a parameter |
| 6 | Data Slider | no | Data Entry MSB | reserved message would write a parameter |
| 92 | A_Undefined92 | no | — | placeholder, nothing binds it |
| 95 | A_Undefined95 | no | — | placeholder, nothing binds it |
| 96 | A_Undefined96 | no | Data Increment | reserved message would write a parameter |
| 103 | A_Undefined103 | no | — | placeholder, nothing binds it |
| 104 | A_Undefined104 | no | — | placeholder, nothing binds it |
| 111 | A_Undefined111 | no | — | placeholder, nothing binds it |
| 120 | A_Undefined120 | no | All Sound Off | reserved message would write a parameter |
| 121 | A_Undefined121 | no | Reset All Controllers | reserved message would write a parameter |
| 123 | All Notes Off | yes | All Notes Off | reserved message would write a parameter |
| 124 | A_Undefined124 | no | Omni Off | reserved message would write a parameter |
| 125 | A_Undefined125 | no | Omni On | reserved message would write a parameter |
| 126 | A_Undefined126 | no | Mono On | reserved message would write a parameter |
| 127 | A_Undefined127 | no | Poly On | reserved message would write a parameter |

### OsTIrus: page A slots left uncovered

| cc | parameter | device accepts it | reserved MIDI meaning | left alone because |
|---|---|---|---|---|
| 0 | Version | no | Bank Select MSB | reserved message would write a parameter |
| 6 | Data Slider | no | Data Entry MSB | reserved message would write a parameter |
| 92 | A_Undefined92 | no | — | placeholder, nothing binds it |
| 95 | A_Undefined95 | no | — | placeholder, nothing binds it |
| 96 | A_Undefined96 | no | Data Increment | reserved message would write a parameter |
| 97 | A_Undefined97 | yes | Data Decrement | reserved message would write a parameter |
| 98 | A_Undefined98 | yes | NRPN LSB | reserved message would write a parameter |
| 99 | A_Undefined99 | yes | NRPN MSB | reserved message would write a parameter |
| 100 | A_Undefined100 | yes | RPN LSB | reserved message would write a parameter |
| 101 | A_Undefined101 | yes | RPN MSB | reserved message would write a parameter |
| 102 | A_Undefined102 | yes | — | placeholder, nothing binds it |
| 120 | A_Undefined120 | no | All Sound Off | reserved message would write a parameter |
| 121 | A_Undefined121 | no | Reset All Controllers | reserved message would write a parameter |
| 123 | All Notes Off | yes | All Notes Off | reserved message would write a parameter |
| 124 | A_Undefined124 | no | Omni Off | reserved message would write a parameter |
| 125 | A_Undefined125 | no | Omni On | reserved message would write a parameter |
| 126 | A_Undefined126 | no | Mono On | reserved message would write a parameter |
| 127 | A_Undefined127 | no | Poly On | reserved message would write a parameter |

### Page B slots the device echoes but the map does not cover

| pp | model | parameter | left alone because |
|---|---|---|---|
| 91 | Osirus | B_Undefined91 | placeholder, nothing binds it |
| 1 | OsTIrus | B_Undefined1 | placeholder, nothing binds it |
| 91 | OsTIrus | B_Undefined91 | placeholder, nothing binds it |

Beyond those, 16 further page B slots are uncovered on Osirus and 13 on OsTIrus, but the device never echoes them and external poly pressure is rejected, so nothing could reach them either way. All are `B_Undefined*` placeholders.

## Testing

Both commits build across all plugins and pass the ctest suite. Several popular third-party skins were scanned to confirm no bound parameter lost its delivery path. Beyond that, an identical set of tests was run against live Osirus and OsTIrus instances, first on `main` and then on this branch, in single and multi mode. The two runs differ in exactly four places, all of them the intended change, and are identical everywhere else.

| | on `main` | on this branch |
|---|---|---|
| `cc 0` → `Version` | Osirus 7→3, OsTIrus 12→3 | dropped |
| `cc 6` → `Data Slider` | 0→99 on both | dropped |
| `cc 123` → `All Notes Off` | 1→0 on both | dropped |
| `cc 17` on channel 5, single mode | writes part 5 | dropped |

Everything that had to stay the same did: `cc 17`, `cc 1` through `cc 4`, `cc 5` on TI, and `cc 122` on Osirus all produce identical changes, host-sourced poly pressure is still rejected, and multi mode is unchanged throughout — `cc 17` on nibble N still reaches part N, nibble 15 still reaches part 15, and with part 1 reassigned to midi channel 6 a control change on nibble 6 moves part 6 while part 1 stays put.

### The controller dump itself

The tests above inject individual control changes, which exercise the live-edit branch of `getPartsForMidiEvent`. The other branch — the device's own reply to `REQUEST_CONTROLLER_DUMP` (`0x37`), which is what the deleted `parseControllerDump` existed to handle — needed covering separately, since that reply is generated inside the device and cannot be reproduced by injecting single messages.

The method relies on a raw sysex reaching the device without refreshing the plugin's parameter cache. A parameter-change message moves a value on the device only, leaving the cache deliberately stale; the controller dump is then the only thing that can resynchronise it, so watching which part catches up reads the addressing decision directly. All parts are seeded with known values first, so a mis-route shows up as the wrong part moving rather than as an ambiguous no-op, and each test asserts the cache did not move before the dump was requested.

Run on Osirus against `a2965f30` and against this branch, and on OsTIrus against this branch:

| Test | Stimulus | Result |
|---|---|---|
| 1 | multi mode, `PAGE_A` write to part 3, dump part 3 | part 3 moves, parts 0-2 and 4-7 unchanged |
| 2 | single mode, `PAGE_A` write to `SINGLE`, dump `SINGLE` | part 0 moves, all other parts unchanged |
| 3 | multi mode, `PAGE_A` write to part 6, dump part 6 | part 6 moves, all other parts unchanged |
| 4 | multi mode, `PAGE_B` write to part 3, dump part 3 | value lands on part 0 |
| 5 | multi mode, `PAGE_B` write to part 0, dump part 0 | part 0 moves, all other parts unchanged |

The Osirus runs are byte-identical between the two builds, controls included. OsTIrus produces the same results on this branch. The dump path behaves exactly as it did before the refactor. Test 4 records a potential pre-existing bug rather than a regression, addressed in #290.

## Interaction between three related PRs

I have submitted three PRs affecting Osirus/Ostirus MIDI channel handling, and they are easiest to reason about together:

- #285 - resolve single mode against the global MIDI channel, instead of a hardcoded 0.
- #290 - emit page B of the controller dump with the part in the status byte's low nibble, instead of a hardcoded 0.
- #291 - route parameter traffic through the shared controller map, with device-sourced traffic exempt from channel resolution.

The table below considers the effects of various combinations of these three PRs on MIDI handling.

- **Behavior 1**: In single mode, if the global MIDI channel is set to channel G != channel 0, and a MIDI-learned mapping is bound to channel G, which part does it resolve to? The same resolution also drives program change routing and the parameter overlay indicators.
- **Behavior 2**: In single mode, if the global MIDI channel is set to channel G != channel 0, and CC is sent to that channel, which part is written?
- **Behavior 3**: In single mode, if the global MIDI channel is set to channel G != channel 0, and CC is sent to channel 0, which part is written?
- **Behavior 4**: In multi mode, if a dump of part N != part 0 is requested, which part do the page B parameters get written to?

Two behaviours are correct in every combination and are left out of the table below: the single-mode controller dump, and the page A half of a multi-mode dump.

| Row | #285 | #290 | #291 | Behavior 1 | Behavior 2 | Behavior 3 | Behavior 4 |
|---|---|---|---|---|---|---|---|
| 1 | no | no | no | no part, wrong | part G, wrong | part 0, wrong | part 0, wrong |
| 2 | no | no | yes | no part, wrong | dropped, wrong | part 0, wrong | part 0, wrong |
| 3 | no | yes | no | no part, wrong | part G, wrong | part 0, wrong | part N, correct |
| 4 | no | yes | yes | no part, wrong | dropped, wrong | part 0, wrong | part N, correct |
| 5 | yes | no | no | part 0, correct | part G, wrong | part 0, wrong | part 0, wrong |
| 6 | yes | no | yes | part 0, correct | part 0, correct | dropped, correct | part 0, wrong |
| 7 | yes | yes | no | part 0, correct | part G, wrong | part 0, wrong | part N, correct |
| 8 | yes | yes | yes | part 0, correct | part 0, correct | dropped, correct | part N, correct |

Row 1 is current behavior, and no row regresses against it, so any subset may be merged in any order. Row 2 is the only one that can look like a regression, since #291 alone drops a control change that is delivered today, but today it is delivered to the wrong part, so neither matches the device.

#290 is independent: it flips Behavior 4 identically in every row. #285 independently fixes Behavior 1. For Behaviors 2 and 3, however, #285 and #291 are complementary rather than independent, and neither alone produces the correct result.

**AI Disclosure: This PR was made with the help of an AI coding assistant.**


ChasonDeshotel:
Nice bit of architectural improvement. I like the direction. Couple thoughts:

1. null-pointer guard is unrelated and silently hides an invalid state.

In `framework/juce/jucePluginLib/controller.cpp`:

```cpp
// This prevents a nullptr dereference caused by swapping between ROMs with different part counts
				if(!param)
					continue;
```
This seems outside the scope of the controller-map refactor and makes the behavioral change a little harder to review. At minimum, I think it should be a separate commit, with logging or an assertion rather than silently swallowing the missing parameter. More broadly, if this condition is caused by switching between ROMs with different part counts, my gut feeling is that another layer should own it—the fact that ROM switching reaches this far into the controller suggests the responsibilities may be a little tangled.

I’ve been doing some adjacent work in #287, extracting TI Control state decoding and preset-file parsing from device and UI code into shared, host-independent components, partly to address this kind of coupling.

2. I really appreciate the extensive manual testing. That said, I think getPartsForMidiEvent() deserves automated coverage. Channel/part/source routing has a lot of edge cases, so automated regression tests would be valuable.

3. Between this PR and #287, I think we should also run one combined regression test: load the captured TI Control state from #287, let the device synchronize the controller, and verify representative page-A and page-B parameters across a few parts.

Neither PR currently owns that end-to-end boundary. I’d put the test with the Virus controller/integration tests rather than in TIControlStateDecoderTest, since #287’s decoder is deliberately host-independent and should remain that way

PR #293, opened 2026-09-04
h8s: skip timer cycles where nothing can happen

charlesvestal:
Note: the below was created heavily via Claude, but the results have been verified both via the test bench and by a human, running on real hardware. 

`Timers::tick()` runs once per instruction and steps every elapsed cycle of every
running channel — roughly 180 iterations per sample per channel, almost all of
them just `tcnt++`. Only the cycles where a channel actually reaches `gra`, `grb`
or wraps to 0 do anything.

This computes the first cycle at which that can happen (`nextEvent`) and, until
then, `tick()` is a single compare; the inner loop jumps straight to the next
event.

### Why it is safe

`timers.tick()` runs **after** `emu.step()`, so the old code updated the counters
to end-of-instruction cycles. `catchUp()` uses `instrStartCycles`, which is
exactly that point, and it runs on any register access — so the CPU can never
observe a stale `tcnt`.

Deferred increments equal per-instruction ones because `inc` is a difference of
floors, and that is additive across any split of the interval.

Two things a reviewer will want to check, called out rather than buried:

- The skip conditions in `update()` and in the `nextEvent` recompute must stay in
  lockstep. If one skips a case the other does not, the horizon is wrong.
- `write()` resets `nextEvent`, because a register write can move the next event
  earlier.

### Measured

Upstream's own `jeTestConsole`, on an idle Raspberry Pi 4B @ 1.8 GHz
(`throttled=0x0`, checked before and after). Counterbalanced B,A,A,B rather than
alternating, because thermal drift plus straight alternation manufactured a
result here once that did not survive:

| position | build | speed | wav bytes |
|----------|-------|-------|-----------|
| 1st | before | 43% | 68,135,468 |
| 2nd | after  | 45% | 71,887,916 |
| 3rd | after  | 45% | 71,518,508 |
| 4th | before | 42% | 67,905,068 |

**≈ +6%.** `after` wins in both middle positions, and `before` reads 43% first
and 42% last, consistent with the 50.6 → 58.4 °C drift across the session. The
wav sizes corroborate from an independent measure: 5.4% more audio rendered in
the same wall clock.

### Bit-exact

All 68,135,424 audio bytes identical.

Worth stating how, because the raw hashes differ and that looks alarming: each
run is stopped on a wall clock, so the faster build simply writes **more** audio.
The comparison is the shorter file as a prefix of the longer:

```
cmp -n $(stat -c %s before.wav) -i 44:44 before.wav after.wav
```

### Scope

Measured on ARM64 (Cortex-A72) only. I have no x86 hardware to test on, and
timing under emulation would be meaningless, so I am not making a claim for x86 —
the change is architecture-neutral, but the number is not.

### DIFF
--- a/source/cpu/h8s/h8s.hpp
+++ b/source/cpu/h8s/h8s.hpp
@@ -51,6 +51,11 @@ public:
 	uint8 exr {0};
 	H8SDevice* maps[1<<24] {};
 	unsigned long long  cycles {0};
+	/* `cycles` as it stood when the current instruction began. Devices that
+	 * defer their per-instruction tick (Timers) catch up to THIS on a register
+	 * access, which is the state the original end-of-previous-instruction tick
+	 * would have left them in. */
+	unsigned long long instrStartCycles {0};
 	unsigned long long pending_irqs {0};
 	
 	enum
@@ -273,6 +278,7 @@ public:
 	uint8 *handle_instr(uint8 *_pc)
 	{
 		pc = _pc;
+		instrStartCycles = cycles;
 		if (execute && pending_irqs)
 		{
 			bool ue = read8(0xfffff2) & 8;
--- a/source/cpu/h8s/h8sdevices.hpp
+++ b/source/cpu/h8s/h8sdevices.hpp
@@ -20,7 +20,18 @@ public:
 	void tick_extclock(int which) {
 		channels[which].gra++;
 	}
+	/* Called once per H8S instruction. Nothing observable happens between
+	 * events (a channel reaching gra, grb or 0), and `nextEvent` is the first
+	 * cycle at which one can, so until then this is one compare. The deferred
+	 * increments are the same as the per-instruction ones because inc is a
+	 * difference of floors, which is additive across any split. Register
+	 * accesses catch up first (`catchUp`), so the H8S never sees a stale tcnt. */
 	void tick()
+	{
+		if (state->cycles < nextEvent) return;
+		update(state->cycles);
+	}
+	void update(unsigned long long now)
 	{
 		for (int i = 0; i < 5; i++)
 		{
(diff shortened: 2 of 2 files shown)


charlesvestal:
**Correction: this is not bit-exact, on program changes. Moving back to draft.**

The bit-exactness claim above is wrong, and the evidence I gave for it was too
short to catch that. I verified 68,135,424 audio bytes — 128.7 s of the factory
demo. The output diverges at **176.0 s**, 47 seconds past where I stopped looking.

Measured on an idle Pi 4B, serial, against a vanilla `main` build of the same
harness, both runs long enough to reach the event:

```
vanilla main vs this branch:  DIFFERS at 93,135,918 bytes = 176.0 s
  span:        739 ms of differing samples
  magnitude:   1.19% of bytes, RMS(diff)/RMS(signal) = 0.00066  (about -64 dB)
  recovery:    none — still differing at 245 s
```

It starts exactly at the demo's patch change to `[5:Feedbacker]`, which is when
the firmware is busiest with timer and status registers. Two vanilla runs are
byte-identical to each other over 378.8 s, so this is the change, not harness
noise.

For what it's worth, the same long comparison clears the other two engine changes
I have open: they are byte-identical to vanilla over 378.7 s and 204.7 s
respectively.

I have not found the cause yet. Ruled out so far: the floor-difference arithmetic
is additive across splits as claimed; the skip conditions in `update()` and the
`nextEvent` recompute are in lockstep; `write()` does reset the schedule;
externally clocked channels are excluded from both loops. The asymmetry I am
looking at now is that upstream raises timer IRQs only from `tick()` after
`step()`, whereas `catchUp()` can raise one mid-instruction from inside a
register read — `pending_irqs` is sampled at instruction start so that ought to
be equivalent, but "ought to be equivalent" is what produced the original claim.

Back to draft until it is either fixed or withdrawn. Sorry for the noise, and
if the answer is obvious to you from the firmware side I would welcome the
pointer.

charlesvestal:
Retracting my "not bit-exact" correction: the 176.0 s divergence is not this PR, and the timer logic was never wrong.

The cause is #300: the ARM64 JIT enters its generated programs with uninitialised registers (the ERAM latch chain among them), so the audio at the demo's 176 s patch change depends on incidental binary layout. ANY change to the tree shifts it — this branch merely produced a different binary. Three builds of unmodified main disagree with each other at the same spot, and the arm64 backend disagrees with the x64 backend there too.

The tick()/catchUp() asymmetry I said I was chasing is innocent — the timer event and IRQ streams were already measured identical between this branch and main while the audio differed, which is exactly the fingerprint of the JIT bug rather than a timing change.

With #300 applied to both sides, this branch is byte-identical to main over 380 s on a Pi 4 (gcc) and 200 s on an M1 (clang), through the 176 s patch change. Once #300 is in, I'll take this out of draft.

PR #294, opened 2026-09-04
esp: dense ARM64 emitter — pre-shifted coefs and a mirrored ring (+65% A72, +45% M1, bit-exact)

charlesvestal:
Note: the below was created heavily via Claude, but the results have been verified both via the test bench and by a human, running on real hardware.

Emits each ESP program as one straight-line block instead of a sequence of
per-op calls, and removes the two things that cost the most inside it.

### The change

**Pre-shifted coefficients.** `shiftAmount` is one of {3,5,6,7} and `coef` is an
`int8`, so `coef << (7 - shiftAmount)` is at most 2032 and fits an `int16`. Then

```
(P * coef) >> shiftAmount   ==   (P * coefShifted) >> 7
```

holds *exactly* under arithmetic shift, so a plain MAC becomes one load and a
fixed immediate shift rather than a variable one. `coefsShifted` is filled beside
`coefs` in `updateCoefEntry()`, which also folds away the duplicated per-core
loop that was there.

**Mirrored iram/gram ring.** The ring becomes a 512-entry buffer addressed as
`iramPos + mem` with no wrap, so the JIT rebases the ring pointer once per call
and reaches every slot with an immediate offset instead of masking every access.
Old-ring slot `Q` lives at `Q+256` while `Q < iramPos`, else at `Q`; `sync()`
moves the single slot whose home changes when `iramPos` decrements, and
`syncShared()` does the same for the shared gram. `ramIdx()` keeps the
interpreter and the host-side accessors on the same index, so the two can never
drift apart.

**`emitOp()` takes `nextIsDmac`.** `last_mulInputA_24` / `last_mulInputB_24` have
exactly one reader — a `kDMAC` — so when the next emitted op of the core is not
one, those writes are skipped. It is computed from the next *statically emitted*
op, which is consistent with `esp_opt.hpp` already declaring jumps unsupported.
The x64 emitter takes the argument and ignores it.

### Why the mirror is `#if defined(__aarch64__)`

Not a placeholder, and not "untested elsewhere", the ring layout has to match
the emitter that generates the code, and the two emitters differ:

- `esp_jit_arm64.cpp` rebases the ring once and addresses slots with immediate
  offsets. It requires the mirrored layout.
- `esp_jit_x64.cpp` emits `and tempB, 0xff` for every access. It requires the
  masked layout.

Since the emitter is selected by architecture, so is the layout. Turning the
mirror on for x86-64 without porting the mirrored addressing into the x64 emitter
would make the JIT and the interpreter index different slots whenever
`iramPos > 0` — incorrect, not merely unmeasured.

With the mirror off, `ramIdx()` reduces to `(offset + iramPos) & IRAM_MASK`,
which is the existing expression character for character; `mirrorFixup()` returns
immediately; and the buffers are 256 entries again. **So x86-64 behaviour is
unchanged**, and the only things it sees are an extra `int16` array in `CoreData`
and the widened `emitOp()` signature. Compile-checked for `x86_64` and `arm64`.

Porting the mirror into the x64 emitter is a reasonable follow-up, and I would
rather do it as its own change with its own measurement than bundle it here on
hardware I do not have.

### Measured

Upstream's own `jeTestConsole`, counterbalanced B,A,A,B at equal wall clock. The
byte totals are the figure to trust — "last interval speed" is a one-second
sample and reads 75% and 66% for two runs whose totals are 0.1% apart.

**Raspberry Pi 4B @ 1.8 GHz, idle (load 0.05–0.2), 52 → 58 °C per run**, 300 s:

| position | build | speed | wav bytes |
|----------|-------|-------|-----------|
| 1st | before | 43% | 68,153,132 |
| 2nd | after  | 75% | 112,570,412 |
| 3rd | after  | 66% | 112,447,532 |
| 4th | before | 43% | 68,008,748 |

Mean ratio **1.653 (+65%)**. `before` reads 43% in both positions 1 and 4 with
totals 0.2% apart, so this is not thermal drift. Corroborated at a second run
length: 900 s runs gave 337,946,156 / 200,435,756 = **1.686**.

**Apple M1**, 90 s runs, to check the change does not cost a wide out-of-order
core anything:

| position | build | speed | wav bytes |
|----------|-------|-------|-----------|
| 1st | before | 390% | 163,292,972 |
| 2nd | after  | 579% | 239,722,796 |
| 3rd | after  | 595% | 235,930,412 |
| 4th | before | 376% | 165,196,844 |

Mean ratio **1.448 (+45%)**. Smaller than the A72's +65%, which is the expected
direction. Disclosure: that machine was *not* idle — it is a working desktop that
carried about one core of steady background load across all four runs.
Counterbalancing at equal wall clock tolerates a steady load; it does not make
the machine idle.

### Bit-exact

**378.7 s of audio byte-identical between the base and this branch on the Pi**,
with a clean same-build self-check over the same window:

```
upstream vs upstream : IDENTICAL over 200,435,712 bytes
upstream vs this PR   : IDENTICAL over 200,435,712 bytes
```

MD5 of the audio payload (`tail -c +45 x.wav | head -c 200435712 | md5sum`) is
`1c01a71ab04edf0f2a4598459fb5b953` for both builds.

Over a shorter 68,153,088-byte window the same payload hashes
`2619badfaaae7642880e41c7a2c1e933` on **Pi/gcc/Linux, M1/clang/macOS, base and
this branch alike** — two architectures, two toolchains, two operating systems,
one hash.

One caveat I would rather state than have found: on the M1, runs of the **same**
binary diverge at 223.8 s, so a reviewer testing on a busy laptop may see a
mismatch there. It is the machine, not the change — six same-build pairs put the
cross-build divergence point at the modal value of their own spread, and the Pi
renders the identical passage deterministically. What sits at that point is a
demo program change.

### Scope

No threading, forking or scheduling changes: only `source/ronaldo/esp/`, and both
measured binaries are single-threaded.

### DIFF
--- a/source/ronaldo/esp/esp.hpp
+++ b/source/ronaldo/esp/esp.hpp
@@ -5,6 +5,7 @@
 #include <algorithm>
 #include <assert.h>
 #include <array>
+#include <string.h>
 
 template <int32_t N> static constexpr int32_t se(int32_t x) { x <<= (32 - N); return x >> (32 - N); }
 
@@ -16,6 +17,14 @@ class ESPCore;
 
 template<int lg2eram_size>
 class ERAM;
+
+// ESP_IRAM_MIRROR: iram/gram are 512-entry buffers addressed as iramPos + mem with NO wrap. This lets
+// the ARM64 JIT rebase the ring pointer once per call and address every slot with an immediate offset.
+// Old-ring physical slot Q lives at Q+256 while Q < iramPos, else at Q; sync() moves the one slot
+// whose home changes when iramPos decrements. Host-side accessors must use the same (unmasked) index.
+#include "esp_jit_types.h"
+#define ESP_IRAM_MIRROR JIT_ARM64
+#define ESP_RAM_BUF_SIZE (ESP_IRAM_MIRROR ? 512 : 256)
 
 #include "esp_opt.hpp"
 
@@ -121,7 +130,7 @@ protected:
 
 template<int lg2eram_size>
 struct SharedState {
-	int32_t gram[256] {};
+	int32_t gram[ESP_RAM_BUF_SIZE] {};
 	uint8_t readback_regs[4] = {0xff, 0xff, 0xff, 0x00};
 	int32_t mulcoeffs[8] = {};
 	ERAM<lg2eram_size> eram;
@@ -144,17 +153,32 @@ public:
 
 	void setup(const uint32_t *_pram, SharedState<lg2eram_size> *_shared) { pram = _pram; shared = _shared; }
 
-	void writeGRAM(int32_t val, uint32_t offset) {shared->gram[(offset + iramPos) & IRAM_MASK] = val;}
-	int32_t readGRAM(uint32_t offset) const {return shared->gram[(offset + iramPos) & IRAM_MASK];}
+	inline uint32_t ramIdx(uint32_t offset) const
+	{
+		return ESP_IRAM_MIRROR ? (offset & IRAM_MASK) + iramPos : (offset + iramPos) & IRAM_MASK;
+	}
 
-	void writeIRAM(int32_t val, uint32_t offset) { iram[(offset + iramPos) & IRAM_MASK] = val; }
-	int32_t readIRAM(uint32_t offset) const {return iram[(offset + iramPos) & IRAM_MASK];}
+	// Called after iramPos has been decremented to newPos (see ESP_IRAM_MIRROR).
+	static inline void mirrorFixup(int32_t* buf, uint32_t newPos)
+	{
+		if (!ESP_IRAM_MIRROR) return;
+		if (newPos == IRAM_MASK) memcpy(&buf[IRAM_SIZE], &buf[0], (IRAM_SIZE - 1) * sizeof(int32_t));
+		else buf[newPos] = buf[newPos + IRAM_SIZE];
+	}
+
+	void writeGRAM(int32_t val, uint32_t offset) {shared->gram[ramIdx(offset)] = val;}
+	int32_t readGRAM(uint32_t offset) const {return shared->gram[ramIdx(offset)];}
+
+	void writeIRAM(int32_t val, uint32_t offset) { iram[ramIdx(offset)] = val; }
+	int32_t readIRAM(uint32_t offset) const {return iram[ramIdx(offset)];}
 
 	void sync() {
 		pc = 0;
 		iramPos = (iramPos - 1) & IRAM_MASK;
+		mirrorFixup(iram, iramPos);
 	}
-	
+	void syncShared() { mirrorFixup(shared->gram, iramPos); }
+
 	void steperam() { if (lg2eram_size) shared->eram.tickCycle((pram[pc] >> 23) & 0x1f, pc); }
 
 	void step() {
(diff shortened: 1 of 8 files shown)


source/ronaldo/esp/esp.hpp
```diff
@@ -17,6 +18,17 @@ class ESPCore;
...
+// the ARM64 JIT rebase the ring pointer once per call and address every slot with an immediate offset.
+// Old-ring physical slot Q lives at Q+256 while Q < iramPos, else at Q; sync() moves the one slot
+// whose home changes when iramPos decrements. Host-side accessors must use the same (unmasked) index.
+#if defined(__aarch64__)
+#define ESP_IRAM_MIRROR 1
+#else
+#define ESP_IRAM_MIRROR 0
+#endif
+#define ESP_RAM_BUF_SIZE (ESP_IRAM_MIRROR ? 512 : 256)
+
```
ChasonDeshotel, line 28: `JIT_ARM64` includes `_M_ARM64`, but `ESP_IRAM_MIRROR` only checks `__aarch64__`. Under MSVC ARM64, this allocates the old 256-entry buffers while the updated ARM64 emitter unconditionally uses the mirrored, rebased addressing scheme and can address up to index 510

If Windows ARM64 is meant to retain the old behavior, the emitter’s rebasing and immediate-offset accesses also need to be conditional on `ESP_IRAM_MIRROR`, falling back to the previous masked addressing. Also suggest deriving `ESP_IRAM_MIRROR` from `JIT_ARM64` in `esp_jit_types.h` so every ARM64 backend uses the mirrored layout

source/ronaldo/esp/esp_opt.hpp
```diff
@@ -223,29 +226,24 @@ class ESPOptimizer
...
-    }
+    uint32_t instr = pram[i];
+    uint32_t op = (instr >> 16) & 0x7c;
+    int8_t coef = se<8>(instr & 0xff);
+    uint32_t shift = (instr >> 8) & 3;
+    uint32_t shiftAmount = (0x3567 >> (shift << 2)) & 0xf;
+    if (op == 0x20 || op == 0x24) shiftAmount = (shift & 1) ? 6 : 7;
+    data.coefs[i] = coef;
+    data.shiftAmounts[i] = shiftAmount;
+    data.coefsShifted[i] = static_cast<int16_t>(static_cast<int32_t>(coef) << (7 - shiftAmount));
```
ChasonDeshotel: `coef` can be negative here, and left-shifting a negative signed value is undefined behavior. Since the intended result fits in `int16_t`, use multiplication instead?

```cpp
data.coefsShifted[i] = static_cast<int16_t>(
    static_cast<int32_t>(coef) * (1 << (7 - shiftAmount)));
```

source/ronaldo/esp/esp.hpp
charlesvestal: Fixed as you suggest: `ESP_IRAM_MIRROR` now derives from `JIT_ARM64` (esp_jit_types.h) so every ARM64 backend gets the 512-entry buffers.

source/ronaldo/esp/esp_opt.hpp
charlesvestal: Fixed with the multiplication form; compiles to the same shift on gcc/clang (verified byte-identical output).

charlesvestal:
Pushed an update; besides the two review fixes, two things changed:

1. The branch is now based on #300 (the ARM64 JIT entry-state fix), so it shows two commits until that lands. This emitter rewrites `jitEnter`/`jitExit`, so it inherits the same obligation to load and store the ESP's persistent state; the previous revision zeroed `last_mulInputA/B_24` per call instead, which disagrees with the x64 backend whenever a program's first op is a DMAC.
2. The `nextIsDmac` store-elision scan now wraps: with entry/exit persistence, the "next emitted op" after a program's last op is the FIRST op of the next call, so a program starting with a DMAC keeps its final `last_mul` values.

Re-verified: serial output byte-identical to main+#300 over 200 s (M1/clang) and as part of the full stack over 380 s (Pi 4/gcc), and the ARM64 output remains byte-identical to the x64 backend. Perf is unchanged — the persistence is ~30 instructions per 384 emulated steps.

PR #295, opened 2026-09-04
esp: recompile only the core whose program words were written (~50% fewer compiles at patch change)

charlesvestal:
A patch load is roughly 60 program write bursts per ASIC, and every one of them
recompiled **both** ESP cores. A core's generated code depends only on its own
program words — core 0 owns intmem words `0..0x3ff`, core 1 the rest, and the
ERAM decode reads core 1's — so `setProgramDirty` now carries a core mask and
`genProgram` compiles only the cores that were actually written.

Both write paths supply the mask. The `0x54` path writes a single word, so it
names one core. The `0x56` path writes `pmem[addr] .. pmem[addr + 4]`, so a burst
that starts below the `0x400` boundary and ends at or above it names both.

The mask accumulates across the three-sample dirty delay and is cleared when the
recompile runs, so several bursts landing in the same window still recompile each
affected core once. `setProgramDirty()` with no argument keeps its old meaning of
"both".

### Measured

The right harness here is a patch sweep, not sustained playback. Eight factory
patch changes across a 10 s render, on an idle Raspberry Pi 4B @ 1.8 GHz
(load 0.08, 46.7 → 52.5 °C), counterbalanced off/on/off/on:

| | core compiles | total compile time | per compile |
|---|---------------|--------------------|-------------|
| without | 1542 | 167.6 / 169.1 ms | 0.109 / 0.110 ms |
| with    |  775 |  83.4 /  83.3 ms | 0.108 ms |

**767 of 1542 core compiles avoided (49.7%)**, and 85 ms of the 168 ms.

Two things worth reading off that table:

- **The per-compile cost is unchanged** (0.108 ms either way). This change removes
  compiles; it does not make them cheaper. So it is independent of anything that
  speeds up code generation itself.
- **The recompile events are identical** — 762 `genProgram` calls in both builds.
  Only the amount of work per event changes, never when it happens.

The mask distribution is the design premise, measured rather than argued: of 762
bursts, **470 touched core 0 alone, 288 touched core 1 alone, and 4 straddled the
boundary** and legitimately needed both.

### What this is and isn't

It is a **stall** fix, not a throughput win. 167.6 ms of compile inside a 10 s
render is 1.7% of wall clock, and it is concentrated at the patch changes rather
than spread across playback. Halving it halves the pause when a player changes
patch. Sustained playback is not measurably faster, and I would rather say that
than have a throughput benchmark say it for me.

### Bit-exact

The sweep renders identically with and without the change — same length, same
SHA-256 — on both an A72 and an Apple M1.

### Disclosure on method

The measurement was made by toggling `m_dirtyCores |= cores` to
`m_dirtyCores = 3` on my own tree, a one-line revert to the previous behaviour so
that nothing but the mask differs. My patch-sweep harness depends on save-state
support that this branch does not carry, so the branch itself was not the binary
benchmarked — the change is identical, but the measurement is one step removed
and I would rather state that plainly.

### DIFF
--- a/source/ronaldo/esp/esp.hpp
+++ b/source/ronaldo/esp/esp.hpp
@@ -409,7 +409,7 @@ public:
 			auto newValue = *reinterpret_cast<uint32_t*>(&intmem[(addr << 2)]);
 
 			if (newValue != oldValue)
-				opt.setProgramDirty();
+				opt.setProgramDirty(addr < 0x400 ? 1 : 2);	// core 0 owns words 0..0x3ff, core 1 the rest
 		}
 		else if (if_mode == 0x55 && (address & 3) == 3) {
 			const int addr = address >> 2;
@@ -463,7 +463,9 @@ public:
 			};
 
 			if (newValues != oldValues)
-				opt.setProgramDirty();
+				// This path writes pmem[addr] .. pmem[addr + 4], so a burst that starts below the
+				// 0x400 boundary and ends at or above it dirties BOTH cores.
+				opt.setProgramDirty((addr < 0x400 ? 1 : 0) | (addr + 4 >= 0x400 ? 2 : 0));
 		}
 		else if (if_mode == 0x57 && (address & 3) == 3) {
 			addr_sel = address & ~3;
--- a/source/ronaldo/esp/esp_opt.hpp
+++ b/source/ronaldo/esp/esp_opt.hpp
@@ -188,29 +188,35 @@ public:
     data_core1.eramPtr = &esp->shared.eram.eram[0];
   }
 
-  void genProgram(ESP<lg2eram_size>* esp)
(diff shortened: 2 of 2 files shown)


charlesvestal:
Withdrawing this one — on reflection it doesn't carry a user-visible benefit worth your review time.

What it does is real but narrow: ~50% fewer ESP core compiles at patch change (1542 -> 775 on a Pi 4B), bit-exact, per-compile cost unchanged. But sustained playback is not measurably faster, and I never demonstrated that the compile burst it removes actually causes an audible dropout — on a typical buffer it's likely absorbed. Rather than have you weigh a 25-line change on a benefit I can't evidence, I'd rather revisit it later alongside the related `a64::Assembler` change, where the combined effect on patch-change compile cost is a story worth telling with one number.

Leaving #293 and #294 as the ones to look at.

PR #297, opened 2026-09-04
je8086: opt-in parallel ASIC pipeline (2.14x on a Pi 4B, bit-exact on a fixed 2-sample delay)

charlesvestal:
Opt-in parallel rendering of the JP-8000's ASIC chain. **Off by default; nothing
changes for any other device.** Opening as a draft because the design question is
genuinely yours — I'd rather be redirected before you spend review on it.

### The idea

The four ASICs are a chain: ASIC1 needs ASIC0's output *for the same sample*, and
so on. A sample therefore cannot be split across cores — which is the usual
objection, and it's correct. But a dependency chain is exactly what pipelining is
for. Stagger them in time instead:

```
stage 0 (H8S + ASIC0)  → sample S
stage 1 (ASIC1)        → sample S-1
stage 2 (ASIC2, ASIC3) → sample S-2
```

Every stage is busy, each on a different sample, and the chain's ordering is
preserved exactly. What crosses between stages is the GRAM handoff the chain
already has — 3, 6 or 8 words depending on the boundary.

### Determinism

Audio is delivered on a **fixed delay**, not when a stage happens to finish: one
sample handed over per sample rendered, taken a constant number of samples late,
so output is tied to counters rather than to thread timing. Bounding how far the
H8S may run ahead is *not* sufficient on its own — three runs at a bounded window
gave three different hashes.

Verified: the pipeline's output is the serial stream delayed by **exactly 2
samples**, byte-identical, at 2, 3 and 4 stages. Shifts of 1, 3 and 4 all fail
immediately, so this is an exact property rather than an approximate one.

One earlier qualification is now resolved: stacked with the engine PRs this
used to diverge at 176 s, which turned out to be #300 — the ARM64 JIT's entry
state, not anything in this branch or the emitter. On top of that fix the
byte-identical property holds for the full stack at 2, 3 and 4 stages,
verified over 380 s on a Pi 4 (gcc) and 200 s on an M1 (clang).

The delay does not grow with stage count — the parent blocks if a stage hasn't
reached the sample yet, so it's a delivery contract rather than pipeline depth.
2 samples at 88.2 kHz internal is ~23 us, and it is **reported, not hidden**:
`getInternalLatencyMidiToOutput()` adds it to the existing 4.5 ms and
`getInternalLatencyInputToOutput()` returns it, both zero unless the pipeline is
running.

### Measured

`jeTestConsole`, idle Raspberry Pi 4B @ 1.8 GHz, equal wall clock, palindrome
order (serial, 3, 2, 2, 3, serial) so drift can't masquerade as a result. The
pipeline is a runtime setting, so this is one binary throughout — no cross-build
baseline to get wrong.

| configuration | vs serial | real-time | run-to-run |
|---------------|-----------|-----------|------------|
| serial | 1.000x | 0.42x | 0.32% |
| 2 stages | 1.575x | 0.64x | 0.01% |
| **3 stages** | **2.143x** | **0.94x** | 0.24% |

So on this hardware it takes the JE-8086 from **0.42x real time, which is
unusable, to 0.94x — borderline**. I want to be precise about that: the pipeline
alone does not clear real time on a Pi with the current engine. It gets there
combined with engine work (#294), and I'd rather say so than let the two get
conflated.

### In combination with the other PRs

Since the standalone 0.94x above is borderline, here is what it does alongside
the three engine PRs (#293, #294, #296). All four merge cleanly onto `main`,
including the two that both touch `h8s.hpp`. Same Pi, same harness, palindromed,
vanilla upstream as the reference:

| build | vs vanilla | real-time |
|-------|-----------|-----------|
| vanilla upstream, serial | 1.000x | 0.43x |
| + #293 #294 #296, serial | 1.83x | 0.83x |
| + this PR, 2 stages | 3.82x | 1.75x |
| + this PR, 3 stages | ~4.5x | ~1.96–2.03x |

So the four together take a Pi 4B from **0.43x to about 2x real time**. The engine
PRs compose almost exactly as their individual claims predict (1.06 × 1.65 ×
1.014 = 1.77 predicted, 1.83 measured), which is a useful independent check that
none of them was double-counting.

Two things worth noting from that run:

- **Composite bit-exactness holds.** With all three engine PRs stacked, serial
  output is byte-identical to vanilla upstream over 68,297,472 bytes.
- **The 176 s divergence reported here earlier is resolved — it was #300**
  (the ARM64 JIT entered its generated programs with uninitialised registers,
  so its output depended on which thread called it; stage threads changed the
  caller). Nothing in this branch's handoff, forwarding or delivery was wrong
  — traces show GRAM handoffs, forwarded register writes and recompile ticks
  byte-identical between serial and pipeline. With #300 in the tree, the full
  stack renders byte-identical to serial at 2, 3 and 4 stages over 380 s on a
  Pi 4 and 200 s on an M1. Details in the comment below.

### Where it does NOT help — please read this before trying it

On an **Apple M1** (4 performance + 4 efficiency cores, busy desktop), same
binary:

| configuration | vs serial |
|---------------|-----------|
| serial | 1.000x |
| 2 stages | 1.873x |
| **4 stages** | **0.333x** |

Four stages is **three times slower than single-threaded**. A pipeline runs at
the pace of its slowest stage, so once stages land on efficiency cores the whole
chain drops to that pace. The same effect, milder, is why 4 stages loses to 3
inside a plugin host even on homogeneous cores: it pins every core and starves
the host's own audio thread.

This is why the count is a user setting rather than something derived from
`hardware_concurrency()`, and why I'd suggest 2-3 as the useful range.

### Honest disclosure

- **A pre-existing nondeterminism, not introduced here.** At the demo's first
  program change (223.8 s in), two runs of the same binary diverge. That happens
  with **stock upstream, single-threaded, no code of mine involved** — verified on
  a byte-identical vanilla build, six pairwise comparisons. It's timing-jitter
  dependent: an idle Pi is deterministic through it in serial (identical over
  378 s) and a busy M1 is not. The pipeline adds jitter, so it surfaces there
  too. I don't know what in the engine is timing-sensitive at a program change,
  and you'll know the firmware far better than I do — happy to open that
  separately if it's news to you.
- A host's per-plugin CPU readout **understates** this badly, because the work
  happens off the audio thread. REAPER showed 3.3% against roughly 1.4 cores.
- Measured on ARM64 only. No x86 hardware here, so no claim for it.

### Shape of the change

Three commits: two small ESP accessors, the pipeline itself, and a device
capability plus the settings control. The framework part follows
`canModifyDspClock()` exactly — the device declares `getMaxDspThreads()`, the
shared DSP & Audio panel shows the control only when the capability is there, and
six of the seven plugins in the tree see no behavioural change at all.

Stages are deliberately not pinned to cores: pinning measured worthless (three
stages, 1.8x either way) and it's what would break a second instance.

The obvious alternative — `clap.thread-pool` — would give this to CLAP hosts
only, and the project ships VST3 and AU as well, so it can't replace the plugin
owning its threads. It could be an additional path later.

### DIFF
--- a/source/framework/juce/jucePluginData/tus_settings_dspaudio.rml
+++ b/source/framework/juce/jucePluginData/tus_settings_dspaudio.rml
@@ -97,6 +97,20 @@
 				</tr>
 			</table>
 		</div>
+		<div id="containerDspThreads" class="settings-advanced">
+			<h1>Multicore DSP (threads)</h1>
+			<settingsspacer1/>
+			<table id="dspThreadsTable" class="settings-table">
+				<tr class="settings-tr" id="dspThreadsEntry" style="display: none;">
+					<td class="settings-td">
+						<div id="dspThreadsButton" class="settings-checkboxwithlabel">
+							<button id="button" class="settings-checkbox"/>
+							<label id="label">Off (default)</label>
+						</div>
+					</td>
+				</tr>
+			</table>
+		</div>
 		<div id="containerDeviceSpecific"/>
 	</div>
 </body>
--- a/source/framework/juce/jucePluginEditorLib/pluginProcessor.cpp
+++ b/source/framework/juce/jucePluginEditorLib/pluginProcessor.cpp
@@ -93,6 +93,17 @@ namespace jucePluginEditorLib
 		return true;
 	}
 
+	bool Processor::setDspThreads(const uint32_t _threads)
+	{
+		if(!pluginLib::Processor::setDspThreads(_threads))
+			return false;
+
+		getConfig().setValue("dspThreads", static_cast<int>(_threads));
+		getConfig().saveIfNeeded();
+
+		return true;
+	}
+
 	bool Processor::hasEditor() const
 	{
 		return true; // (change this to false if you choose to not supply an editor)
--- a/source/framework/juce/jucePluginEditorLib/pluginProcessor.h
+++ b/source/framework/juce/jucePluginEditorLib/pluginProcessor.h
@@ -19,6 +19,7 @@ namespace jucePluginEditorLib
 		juce::PropertiesFile& getConfig() { return m_config; }
 
 		bool setLatencyBlocks(uint32_t _blocks) override;
+		bool setDspThreads(uint32_t _threads) override;
 
 		bool hasEditor() const override;
 		juce::AudioProcessorEditor* createEditor() override;
--- a/source/framework/juce/jucePluginEditorLib/settingsDspAudio.cpp
+++ b/source/framework/juce/jucePluginEditorLib/settingsDspAudio.cpp
@@ -142,6 +142,71 @@ namespace jucePluginEditorLib
 			}
 		}
 
+		// Multicore DSP buttons. Only offered by devices that can spread their DSP
+		// over several threads; everything else keeps the container hidden.
+		{
+			const auto maxThreads = m_processor.getMaxDspThreads();
+
+			if (maxThreads > 1)
+			{
+				if (auto* threadsTemplate = juceRmlUi::helper::findChild(_root, "dspThreadsEntry", false))
+				{
+					const auto current = m_processor.getDspThreads();
+					auto* parent = threadsTemplate->GetParentNode();
+
+					for (uint32_t threads = 0; threads <= maxThreads; ++threads)
+					{
+						if (threads == 1)	// one thread is the same thing as Off
+							continue;
+
+						auto clone = threadsTemplate->Clone();
+						auto* clonePtr = clone.get();
+						clonePtr->SetId("");
+						clonePtr->RemoveProperty("display");
+
+						auto* buttonContainer = juceRmlUi::helper::findChild(clonePtr, "dspThreadsButton");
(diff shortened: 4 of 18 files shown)


charlesvestal:
The 176 s qualification in the description is resolved, and the answer is better than I hoped: nothing in this branch or in the emitter branch is wrong. The cause is #300 — the ARM64 JIT enters its generated programs with uninitialised registers, so its output depends on WHO CALLS IT. Running ASICs from stage threads changes the caller; so does merely building a different binary, which is why the "interaction" pointed at the emitter. The mirrored-ring lead I described was wrong: traces show the GRAM handoffs, forwarded register writes and recompile ticks all byte-identical between serial and pipeline — the stray write was a single ERAM store from an uninitialised latch register during boot, surfacing 172 s later at the patch change.

With #300 in the tree, the full stack (this branch + the emitter + the H8S changes) renders byte-identical to serial at 2, 3 and 4 stages over 380 s on a Pi 4 and 200 s on an M1, through the 176 s patch change. So the delivery contract holds unqualified once that fix is present; I've updated the description accordingly and will rebase the claim on #300 when it lands.

PR #300, opened 2026-09-05
esp: ARM64 JIT must load and store the ESP's persistent state, not inherit registers

charlesvestal:
The ESP's ERAM latch chain (`eramReadLatch`, `eramWriteLatch/Next`,
`eramVarOffset`, `eramEffectiveAddr`), the DMAC inputs (`last_mulInputA/B_24`)
and the six accumulators persist across samples: the interpreter keeps them in
the ESP object, and the x64 backend loads and stores them through the
`JitInputData` pointers. The ARM64 backend kept them purely in registers and
never initialised them — on entry to the generated program each held whatever
the calling code happened to leave in that register.

With a fixed calling sequence that garbage is reproducible, which is why this
never showed: one binary's output is stable run to run. But it is an accident
of code generation. Three builds of the same source (different incidental code
around `callOptimized`) each produce different audio from the factory demo's
patch change to `5:Feedbacker` at ~176 s — a stray ERAM write during boot that
surfaces there as ~-115 dB noise for a second or two. Anything that changes
the caller changes the output: another compiler, an unrelated PR in the tree,
or running an ASIC from a worker thread.

This fix loads the persistent state in `jitEnter` and writes it back in
`jitExit`, matching the x64 backend's contract; registers with no backing
state are zeroed so entry state is deterministic. Cost is ~30 instructions per
call, i.e. once per 384 emulated ESP steps — not measurable in the render
benches here.

After this change, the ARM64 backend's output is **byte-identical to the x64
backend's** over a 200 s demo render (fat binary on an M1, x64 slice under
Rosetta) — before it, the two differ at 176 s. It also makes the demo render
invariant across builds, threads and compilers on arm64 (verified 200 s on
M1/clang, 380 s on a Pi 4/gcc).

This is also the cause of the divergence I reported against #293 — see the
correction there.

### DIFF
--- a/source/ronaldo/esp/esp_jit_arm64.cpp
+++ b/source/ronaldo/esp/esp_jit_arm64.cpp
@@ -76,7 +76,7 @@ namespace esp
 	constexpr auto mulInA = x27;
 	constexpr auto mulInB = x28;
 	
-	EspJitArm64::EspJitArm64(Asm& a, const JitInputData&) : m_asm(a)
+	EspJitArm64::EspJitArm64(Asm& a, const JitInputData& _data) : m_asm(a), m_data(_data)
 	{
 	}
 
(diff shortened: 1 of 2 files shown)


charlesvestal:
For context on how this surfaced: #297 — the opt-in parallel ASIC pipeline — is up as a draft if you want to see the approach. Verifying its "byte-identical, fixed 2-sample delay" claim is what found this bug: running the ASICs from worker threads changes which code calls the JIT'd programs, and the output changed with it, even though every input to the ASICs (GRAM handoffs, forwarded register writes, recompile timing) traced identical between serial and pipeline.

That divergence, the one I wrongly pinned on #293, and an apparent #294↔#297 interaction all reduce to this one issue. With this fix in the tree, the ARM64 backend matches the x64 backend byte-for-byte, and the full stack (#293 + #294 + #296 + #297) renders byte-identical to serial at 2, 3 and 4 pipeline stages over 380 s on a Pi 4 and 200 s on an M1.

PR #304, opened 2026-09-11
88emuplayer: render audio synchronously in processBlock

masanaohayashi:
This PR moves emulation processing from the worker thread into `processBlock()` and removes the extra buffering latency.

The worker keeps the audio callback responsive, but it cannot prevent dropouts if emulation cannot keep up. It also moves most of the processing cost outside the callback, where it may not be reflected accurately in the DAW’s audio performance meter.

Synchronous rendering lets the host account for the processing cost more directly and avoids the worker queue’s added latency.

Validation covers calling-thread execution, MIDI timing, zero and variable block sizes, and playback on both MIDI ports at 32, 44.1, and 48 kHz.


giulioz:
Thank you! Just merged manually into the next 2.2.21 version.

PR #306, opened 2026-09-22
88emu: fix windows build

sukerokushin:
I kept getting `MessageBoxA` errors while building from source (VS2022, just like github actions uses) and after some research i found out JUCE was adding conflicting headers. undefining MessageBox in Emu88Standalone fixed the build.

this is a very script-kiddie oneliner fix and i dont know cpp dev so i strongly encourage reviews


mrhso:
https://github.com/dsp56300/gearmulator/commit/d0b9460daff775bbda4395ea173b75f8b8198a01

PR #307, opened 2026-09-22
88emu: fix the FFmpeg command execution for Windows

mrhso:


mrhso:
Actually, I think there are still some issues, especially when the input path contains non-ASCII characters. Should using the `W` functions, such as `_wpopen`, instead?

giulioz:
Good point, will fix

